PR #175, opened 2020-03-13
fixed advancedSearch bug

hoseinmirian:


terazus (changes requested):
- [ ] Missing some unit tests (need 100% coverage).
- [ ] Missing code documentation.
- [x] Reset button should reset all filters. It does not anymore (see https://dev--fairsharing.netlify.com/ behaviour)
The autocomplete fields should be redesigned pretty soon anyway so this is not useful ATM.

src/components/Records/SearchFilters.vue
```diff
@@ -76,68 +77,113 @@
...
+            cleanStrings: function (stringArray) {
+                let output = [];
+                const _module = this;
+                stringArray.forEach(function (string) {
+                    output.push(_module.cleanString(string))
+                });
+                return output;
+            },
+            computedKey: function (filter, key) {
+                if (filter.hasOwnProperty('key')) {
```
terazus, line 170: Replace with Object.prototype.hasOwnProperty.call(objectName, fieldName)

terazus (changes requested):

src/components/Records/SearchFilters.vue
hoseinmirian: ok thanks.

PR #177, opened 2020-03-16
Redirect search registry pages

hoseinmirian:
Code works fine but one test => "can get the records" is not passed.


hoseinmirian:
changed branch due to updated dev

PR #201, opened 2020-03-25
adding Throttle to pagination completed and tested

hoseinmirian:


src/components/Records/Pagination.vue
```diff
@@ -43,80 +43,88 @@
...
+            });
+            currentQuery.page = pageNumber;
+            await _module.$router.push({
+              name: _module.$route.name,
+              query: currentQuery
+            });
+          }
+        }
+      },
+      PaginatePermission: throttle(function () { this.allowPaginate=true }, 1200),
```
terazus, line 112: Have you tried wrapping throttle around paginate() instead? It's easier to write/read but make the tests much harder.
No changed required. Just a consideration.

src/components/Records/tests/Pagination.spec.js
```diff
@@ -114,5 +105,25 @@ describe("Pagination.vue", () => {
     });
 
 
+    it('Has a paginate() method that sets the current page in the URL query', () => {
```
terazus, line 102: You cheat here but setting up wrapper.vm.currentPage manually.

src/components/Records/tests/Pagination.spec.js
```diff
@@ -55,44 +62,28 @@ describe("Pagination.vue", () => {
...
-        expect(wrapper.vm.currentPage).toBe(3);
-        wrapper.vm.paginate(3);
-        expect(wrapper.vm.currentPage).toBe(3);
-        expect(wrapper.vm.$route.name).toBe("Standards");
-        expect(wrapper.vm.$route.query.page).toBe("3");
-    });
 
     it('Has a first() method that always redirect to the first page', () => {
 
+
```
terazus: Please clean up a bit the spaces.

src/components/Records/tests/Pagination.spec.js
```diff
@@ -114,5 +105,25 @@ describe("Pagination.vue", () => {
     });
 
 
+    it('Has a paginate() method that sets the current page in the URL query', () => {
+
+        $route.query = {page: "12"};
+        $route.name = "Standards";
+        anotherWrapper2 = shallowMount(Pagination, {
```
terazus: declared but unused

src/components/Records/tests/Pagination.spec.js
```diff
@@ -114,5 +105,25 @@ describe("Pagination.vue", () => {
...
+            mocks: {$route, $router},
+            propsData: {
+                totalPages: 10,
+                default: 0
+            }
+        });
+
+        wrapper.vm.$route.query.page = "12";
+        wrapper.vm.currentPage = 2;
+        wrapper.vm.paginate(12);
```
terazus: paginate() is called but due to throttle the permission method will prevent the code from executing.
This is why paginate will not set the correct wrapper.vm.currentPage

Same for you last() method test. The code inside paginate() is only called for the first() method test. Then, paginate() in all subsequent tests will be prevented by throttle.

src/components/Records/Pagination.vue
```diff
@@ -43,80 +43,88 @@
...
+        allowPaginate: true
+      }
+    },
+    watch: {
+      '$route.name': function () {
+        this.currentPage = 1;
+      },
+      '$route.query': function (newVal) {
+        let _module = this;
+        if (!Object.prototype.hasOwnProperty.call(newVal, "page")) {
```
terazus, line 73: I recently figured out an easier wait of checking for properties:
Object.keys(myVar).includes("myField").
Im not asking for a change. This is just an informative note.

src/components/Records/tests/Pagination.spec.js
```diff
@@ -55,44 +62,28 @@ describe("Pagination.vue", () => {
...
         wrapper.vm.currentPage = 10;
         wrapper.vm.$route.query.page = "1";
 
+        wrapper.vm.allowPaginate = true;
         wrapper.vm.$router.push = push;
         wrapper.vm.first();
         wrapper.vm.$route.query.page = "1";
-
         expect(wrapper.vm.currentPage).toBe(1);
         expect(wrapper.vm.$route.query.page).toBe("1");
```
terazus: This is not very usefull since you go wrapper.vm.$router.query.page = "1" two lines above.
Replace it with:          expect($router.push).toHaveBeenCalledTimes(1);
This way we tests that $router.push has been called the first time, and not the second (see line 93).

terazus (commented):
Please read the comments.

src/components/Records/Pagination.vue
hoseinmirian: this could be a good propose but due to the temporary behavior of Throttle wouldn't be too much work while we know it will be gone later on? I think we can improve it but atm its fine,

src/components/Records/tests/Pagination.spec.js
hoseinmirian: what should I do?

src/components/Records/tests/Pagination.spec.js
hoseinmirian: done

src/components/Records/Pagination.vue
hoseinmirian: that's better and newer. I changed it to include.

src/components/Records/tests/Pagination.spec.js
hoseinmirian: its done

src/components/Records/tests/Pagination.spec.js
hoseinmirian: I fixed it by adding a isTesting variable in the root and make sure in testing it will make sure is allowed for paginate

src/components/Records/tests/Pagination.spec.js
hoseinmirian: removed

src/components/Records/Pagination.vue
hoseinmirian: I did change it and it caused some problem so ATM I stuck with the same old method.

PR #218, opened 2020-04-20
Dom oauth

terazus:


hoseinmirian (commented):
line 130 show1 in login.vue. what is its use?

PR #310, opened 2020-07-15
uiController test coverage reached 100 percent

hoseinmirian:


terazus (changes requested):
the test "can check the toggleOverflow" needs to be moved to a separate file in /test/units/store/UiControllerStore.spec.js
the test also needs one or several expect() method to be valid.

PR #318, opened 2020-07-21
Layout refactor 316

knirirr:
Refactoring mostly complete, other than the possibility of moving associatedRecords into a utils file.


hoseinmirian:
@knirirr how about changing the name of RecordsMixin to something more generic because we might use it in different places in the future. Also, the way mixin is been used globally is addressed on the below link. I have done one version in VuetifyPlayground repo. 

https://github.com/FAIRsharing/vuetifyPlayground/blob/master/src/main.js
Thanks.

hoseinmirian (changes requested):
@knirirr how about changing the name of RecordsMixin to something more generic because we might use it in different places in the future. it has the advantage of being called once globally and we won't need to import it each time. Also, the way mixin is been used globally is addressed on the below link. I have done one version in VuetifyPlayground repo.

initialized at: https://github.com/FAIRsharing/vuetifyPlayground/blob/master/src/main.js
called at: https://github.com/FAIRsharing/vuetifyPlayground/blob/master/src/App.vue
Thanks.

knirirr:
What name would you consider more appropriate, and can you think of any other location where it might be required?
It seems strange to me to import modules etc. globally when they are only required in some areas of the code. Is there a particular reason to do this? I see you have mentioned having to import it each time, but I wouldn't have thought this much of an imposition.

terazus:
You can rename it to stringsUtils.js and we can keep all string shared related methods/filters there (since it's actually what's it's doingt atm)

hoseinmirian:
> 
> 
> What name would you consider more appropriate, and can you think of any other location where it might be required?
> It seems strange to me to import modules etc. globally when they are only required in some areas of the code. Is there a particular reason to do this? I see you have mentioned having to import it each time, but I wouldn't have thought this much of an imposition.

I and Dom decided once to have a global function for repeated actions such as you have done like capitalize, etc. The reason I am saying is that like now,you have written a line import ...mixin from ../ whatever. when we globally import it we won't need to call import each time we want to use it in our components. I suggest name it for now something like "global_mixin"

hoseinmirian:
> 
> 
> You can rename it to stringsUtils.js and we can keep all string shared related methods/filters there (since it's actually what's it's doingt atm)

@knirirr its a good name.

hoseinmirian:
@knirirr please also notice that recordStatus is having a problem showing the status of the item(circle shape). I have attached a screenshot
image

knirirr:
@hoseinmirian can you supply a bit more information on what that error means, please?

terazus:
> 
> 
> @hoseinmirian can you supply a bit more information on what that error means, please?

I can answer that: it means that the record.type is not correctly received by the RecordStatus.vue component.

EDIT:
I tracked the problem down to here: https://github.com/FAIRsharing/fairsharing.github.io/blob/layout_refactor_316/src/components/Records/Record/GeneralInfo.vue#L28
=> should be `<record-status :record="currentRecord['fairsharingRecord']" />`

hoseinmirian:
@knirirr also please change defaultValue to default. this is the way vue understands the default value. 
image

hoseinmirian:
@knirirr the structuring of getters is very nice. please also remember to activate esLint indent before pushing.,, as the webstorm warns its not been followed. just press ctrl+enter on the yellow area and select the second item. also make sure eslint is activated. 
for css and script code reformating use alt+ctrl+L and then use the alt+enter to format it the tags in html to eslint-friendly
image

hoseinmirian:
also, there is a small css problem here that has been overlooked. the links are not activated in my version as default when the mouse hovers they will become underlined while in your version the links are already underlined. have attached my version
image

knirirr:
Various changes have been pushed.
I note that the font-awesome Twitter icon doesn't appear in the footer. This also seems to be the case in the new-layout-design branch, and I have not been able to find out why. 
N.B. all tests run locally but one fails when run by Github.

hoseinmirian:
@knirirr in record number 9 we have 2 organizations with the same name in the back-end which causes problem in front-end. if this is a bug which will be fixed by migration then theres no need to touch the front end code. The problem happens cz each dynamic tag has a key and key is named by the the title. the rule is to not have same key for each generated tag. 

image

knirirr:
> @knirirr in record number 9 we have 2 organizations with the same name in the back-end which causes problem in front-end. if this is a bug which will be fixed by migration then theres no need to touch the front end code. The problem happens cz each dynamic tag has a key and key is named by the the title. the rule is to not have same key for each generated tag.

Probably a migration problem. This would be a matter for @ramGranell to investigate.

hoseinmirian:
@knirirr code has not touched any other places except where it was supposed to but the tests are not passing in the server while they are passed completely in the local. @terazus would you please tell us if there is any odd you might guess. Thanks.

terazus:
@all very likely, if the tests are passing locally but not remotely, a mock issue with an API call.

ramGranell:
Hi, I was just checking the migrated data and I couldn't find that there are two organisations with the same name "United States Department of Agriculture, Agricultural Research Service (USDA-ARS)" .The organisations that contain "USDA-ARS" are these four:
-United States Department of Agriculture, Agricultural Research Service (USDA-ARS)
-United States Department of Agriculture, Agricultural Research Service (USDA-ARS) Beltsville Agricultural Research Center, Beltsville, USA
-United States Department of Agriculture, Agricultural Research Service (USDA-ARS) Corn Insects and Crop Genetics Research Unit, Indiana, USA
-United States Department of Agriculture, Agricultural Research Service (USDA-ARS) Crop Germplasm Research Unit, College Station, USA

@knirirr , if you want to check in the database that they are using is just checking: Organisation.where("name like ?", "%USDA-ARS%") 

is it possible that there are two links between the record and the same organisation? Can you give me information about the record?

hoseinmirian:
> 
> 
> Hi, I was just checking the migrated data and I couldn't find that there are two organisations with the same name "United States Department of Agriculture, Agricultural Research Service (USDA-ARS)" .The organisations that contain "USDA-ARS" are these four:
> -United States Department of Agriculture, Agricultural Research Service (USDA-ARS)
> -United States Department of Agriculture, Agricultural Research Service (USDA-ARS) Beltsville Agricultural Research Center, Beltsville, USA
> -United States Department of Agriculture, Agricultural Research Service (USDA-ARS) Corn Insects and Crop Genetics Research Unit, Indiana, USA
> -United States Department of Agriculture, Agricultural Research Service (USDA-ARS) Crop Germplasm Research Unit, College Station, USA
> 
> @knirirr , if you want to check in the database that they are using is just checking: Organisation.where("name like ?", "%USDA-ARS%")
> 
> is it possible that there are two links between the record and the same organisation? Can you give me information about the record?

@ramGranell this is the json of the data I receive in the picture above . as you can see in record No.9 there are 2 repeated data received as organization  node. 

image

hoseinmirian:
> 
> 
> @ALL very likely, if the tests are passing locally but not remotely, a mock issue with an API call.

@knirirr do we have any mock somewhere new? as far as I checked you did only the initialization tests. Please also review the new changes I pushed.

hoseinmirian:
> 
> 
> @ALL very likely, if the tests are passing locally but not remotely, a mock issue with an API call.

@terazus the problem is with Editing systems code which is weird. 

@hoseinmirian no it's not :) please read the latest build failures more in details ! Only one component of the edit system is affected (editPublications), and many other components not related to edit are affected also (Record/Records/JumpToTop/...)

ramGranell:
> > Hi, I was just checking the migrated data and I couldn't find that there are two organisations with the same name "United States Department of Agriculture, Agricultural Research Service (USDA-ARS)" .The organisations that contain "USDA-ARS" are these four:
> > -United States Department of Agriculture, Agricultural Research Service (USDA-ARS)
> > -United States Department of Agriculture, Agricultural Research Service (USDA-ARS) Beltsville Agricultural Research Center, Beltsville, USA
> > -United States Department of Agriculture, Agricultural Research Service (USDA-ARS) Corn Insects and Crop Genetics Research Unit, Indiana, USA
> > -United States Department of Agriculture, Agricultural Research Service (USDA-ARS) Crop Germplasm Research Unit, College Station, USA
> > @knirirr , if you want to check in the database that they are using is just checking: Organisation.where("name like ?", "%USDA-ARS%")
> > is it possible that there are two links between the record and the same organisation? Can you give me information about the record?
> 
> @ramGranell this is the json of the data I receive in the picture above . as you can see in record No.9 there are 2 repeated data received as organization node.
> 
> image

Ok, I saw what happened. Data is correct, the issue that the database allows to have two or more links between the same organisation and record (OrganisationLink) as it was the case. The difference is that the field "relation" in OrganisationLink is different for each one of the entries in OraganisationLink, in this case one is "funds" and the other "maintains"

terazus:
> Ok, I saw what happened. Data is correct, the issue that the database allows to have two or more links between the same organisation and record (OrganisationLink) as it was the case. The difference is that the field "relation" in OrganisationLink is different for each one of the entries in OraganisationLink, in this case one is "funds" and the other "maintains"

@hoseinmirian 
=> this is unrelated to this PR anyway but the solution is here: https://github.com/FAIRsharing/fairsharing.github.io/blob/master/src/components/GraphClient/queries/getRecord.json#L86
You need to change **organisations** with **organisationLinks** and look at the graphQL documentation for the available fields. This will allow you to get the labels @ramGranell talked about (funds/maintains/...)

hoseinmirian:
> documentation

I can not see any organisationLinks here. 
https://api.fairsharing.org/schema/fairsharingrecord.doc.html
@knirirr would you please help me with this task. Thanks

This is what I get by calling fairsharingRecord
image

knirirr:
@hoseinmirian I think this will need a separate ticket. I will open one.

hoseinmirian:
> 
> 
> @hoseinmirian I think this will need a separate ticket. I will open one.

ok then I merge the current branch and I make a new ticket and assign it to both of us, Thanks

PR #325, opened 2020-07-22
Hossein changing login popup style

hoseinmirian:


src/views/Users/Login/Login.vue
```diff
@@ -11,151 +11,146 @@
...
+
     /** This component handles the login page
      *
      */
     export default {
         name: "Login",
         components: {MessageHandler},
         filters: {
-            capitalize: function(value){
+            capitalize: function (value) {
```
terazus, line 115: use stringUtils for capitalize plz :)

src/views/Users/Login/Login.vue
```diff
@@ -164,31 +159,30 @@
         },
         methods: {
             ...mapActions('users', ['login', 'logout']),
-            logUser: async function(){
+            logUser: async function () {
```
terazus, line 157: there's another much more stylish way of writing in case you are interested. It would be great if we could harmonize all the methods.
` async logUser() { 
// code goes here
} ;`
I'm not expecting a modification here.

terazus (changes requested):

src/views/Users/Login/Login.vue
```diff
@@ -11,151 +11,146 @@
...
       </v-row>
     </v-container>
   </v-layout>
 </template>
 
 <script>
-    import { mapActions, mapState } from 'vuex'
+    import {mapActions, mapState} from 'vuex'
     import MessageHandler from "../../../components/Users/MessageHandler";
+
```
terazus, line 107: Im adding a review !

terazus (changes requested):

src/views/Users/Login/Login.vue
hoseinmirian: @terazus I ll do all the changes you need in separate tickets later. this ticket only focus on the style.

src/views/Users/Login/Login.vue
```diff
@@ -11,151 +11,141 @@
...
         </v-col>
       </v-row>
     </v-container>
   </v-layout>
 </template>
 
 <script>
-    import { mapActions, mapState } from 'vuex'
+    import {mapActions, mapState} from 'vuex'
     import MessageHandler from "../../../components/Users/MessageHandler";
```
terazus, line 106: We can change ../../../ with @/

terazus (approved):

PR #334, opened 2020-07-23
Hossein hiding menu after redirecting

terazus:


knirirr:
I've added one small test.

PR #351, opened 2020-07-28
Dom search query

terazus:


knirirr (changes requested):
Other than the close button looking a bit wonky, which I don't know how to fix, It looks good to me.

Screenshot 2020-07-28 at 13 43 13

terazus:
I think this is a CSS issue with all the icons inside fields. The autocomplete from the search panels are also affected.
I think i already addressed this with @hoseinmirian. Do you have a ticket for that hossein?

hoseinmirian:
the reason I used prevent for submitting form is a cliche issue addressed in the below link. @terazus @knirirr 
https://laracasts.com/discuss/channels/vue/atsubmitprevent-or-just-atclick?page=1

hoseinmirian (approved):
now its fine

PR #359, opened 2020-07-29
"No data found" added in Records.vue if one result is returned. 

### LINKED ISSUE #355: scroll issue in records if one or no records found

hoseinmirian:
state 1: one result is back and state 2: no result is back. images are attached. Please only focus on the task purpose which is the title of the ticket. 

state1
state2


knirirr:
I'd prefer text along the lines of "No records match your search", or similar, but it's not particularly important and what's there does at least explain the pur

knirirr (approved):

src/components/Records/Search/Output/SearchOutput.vue
```diff
@@ -15,7 +15,10 @@
     />
 
     <!--List Row-->
-    <div :class="['opacity-0-transition',{'opacity-1-transition':!isColumnList}]">
+    <div
+      v-if="records.length"
```
terazus: why not v-if="loading" ?

src/components/Records/Search/Output/SearchOutput.vue
```diff
@@ -33,15 +36,31 @@
...
+        icon="info"
+      >
+        No Data Found!
+      </v-alert>
+    </div>
 
     <!-- Card view -->
-    <div :class="['opacity-0-transition',{'opacity-1-transition':isColumnList}]">
+    <div
+      v-if="records.length"
```
terazus: same as above

terazus (changes requested):

src/components/Records/Search/Output/SearchOutput.vue
hoseinmirian: cz this if is about checking whether there is no data available not related to loading data

src/components/Records/Search/Output/SearchOutput.vue
hoseinmirian: same answer as above

hoseinmirian:
> 
> 
> I'd prefer text along the lines of "No records match your search", or similar, but it's not particularly important and what's there does at least explain the pur

gona update it now

src/components/Records/Search/Output/SearchOutput.vue
terazus: OK i understand.
I think making a small computed property that either returns null or the records would make more sense in terms of logic. You can then do v-if='!records' for instance instead of testing for an existing length then testing for a length inferior to 1.
Alternativaly, we can make a getter in the records store that would do the same thing: return the results of the search, or null.

terazus (changes requested):
The first v-if checks for records.length and the v-else-if checks for records.length < 1 => harmonize this plz.

My solution:
If you can't harmonize the checks because of the data structure in the store, i suggest we change records to come from a getter that either returns an array or null

hoseinmirian:
Added getter to read the records length

tests/unit/views/Records/Records.spec.js
```diff
@@ -213,29 +213,29 @@ describe("Records.vue", () => {
     })
 
 
-    it("can check responsiveClassObject",  () => {
+    it("can check responsiveClassObject", () => {
```
terazus, line 217: missing expect statement.

tests/unit/views/Records/Records.spec.js
```diff
@@ -213,29 +213,29 @@ describe("Records.vue", () => {
...
         wrapper.vm.stickToLeft = true
         vuetify.framework.breakpoint.xlOnly = true
     })
 
-    it("can onScroll function work properly",  () => {
+    it("can onScroll function work properly", () => {
```
terazus, line 228: missing expect statements.

terazus (changes requested):
Missing expect statements, otherwise its ok.

tests/unit/views/Records/Records.spec.js
terazus: good !

tests/unit/views/Records/Records.spec.js
```diff
@@ -213,29 +214,40 @@ describe("Records.vue", () => {
...
+        wrapper.vm.$store.state.records.records = ['1', '2', '3'];
         wrapper.vm.offsetTop = 150;
         let mEvent = {
-            target: { scrollTop:150},
+            target: {scrollTop: 150},
         };
+        actions.commit = jest.fn();
+        actions.setGeneralUIAttributesAction({})
         wrapper.vm.onScroll(mEvent);
+        expect(actions.commit).toHaveBeenCalled();
```
terazus: use toHaveBeenCalledTimes() and toHaveBeenCalledWith() instead ^^

terazus (changes requested):

terazus (approved):

Issue #371, opened 2020-08-04
Register button bug


terazus:
When trying to login, if the component is not displayed as a pop up then it is not working.
The method tries to call closePopUp which doesn't exist on the component when it's displayed as a page.

PR #378, opened 2020-08-05
Hossein fixing reigster button bug #371

hoseinmirian:


terazus (changes requested):
Unfortunatly the problem is still persistant:
bug

PR #379, opened 2020-08-05
fixed bug with register page and login button closes: #371

### LINKED ISSUE #371: Register button bug

hoseinmirian:


terazus (dismissed):

terazus (commented):
@hoseinmirian Please review my latest commits

terazus (approved):

PR #382, opened 2020-08-07
Hossein making chips holder sticky #298

hoseinmirian:


terazus (changes requested):
The filter Chip zone is now duplicated and appears before the page banner and in the output zone:
image

terazus (approved):

PR #420, opened 2020-08-18
moved searchBox to leftPanel

### LINKED ISSUE #385: Move search bar
On the search pages the search bar should be the very first filter in the left pannel.
Separate it from the other filters (maybe with an <hr> => <v-divider> in vuetify).

hoseinmirian:


knirirr (approved):

ramGranell (approved):
It seems to work properly!

terazus (changes requested):
src/components/Records/Search/Input/StringSearch.vue => we want to make this component a little more modular. We are going to reuse it in the header and we want to customize it's width and it's placeholder text. Maybe use a property?

src/components/Navigation/Header.vue
```diff
@@ -12,78 +12,83 @@
...
-    <nav>
-      <ul
-        v-if="!$vuetify.breakpoint.sm && !$vuetify.breakpoint.xs"
-        class="d-flex flex-row align-center flex-wrap"
-      >
-        <li
-          v-for="(item, itemIndex) in links"
-          :key="'navBarTopMenuItem_' + itemIndex"
+    <div
+      id="custom-width"
```
terazus: Use a class rather than an ID plz.
ID should be meaningful and describe what the item is about semantically speaking, not a means to access CSS properties :)

src/components/Records/Search/Input/SearchInput.vue
```diff
@@ -13,6 +13,17 @@
       <h2 class="d-none">
         Filter List
       </h2>
+
+      <!-- Search Box -->
+      <string-search
+        place-holder="Search through all data."
```
terazus: placeholder is one word :)

src/components/Records/Search/Input/StringSearch.vue
```diff
@@ -1,36 +1,60 @@
...
+      <span class="button-text-size">Search</span>
     </v-btn>
   </form>
 </template>
 
 <script>
 export default {
   name: "StringSearch",
+  props: {
+    placeHolder: {default: "", type: String}
```
terazus: use placeholder instead of placeHolder so you can have a single word.

terazus (changes requested):

src/components/Records/Search/Input/SearchInput.vue
hoseinmirian: this is vue convention. when u use a capital letter in the names, it will chunk it down to dash-style when used as tags

src/components/Records/Search/Input/StringSearch.vue
hoseinmirian: I disagree. please read this link. this is vue convention. 
https://vuejs.org/v2/guide/components-props.html

terazus (approved):

PR #422, opened 2020-08-18
Mt keyword links 395

### LINKED ISSUE #395: Records keywords should behave as search terms
On the search results, users shouild be able to click each keyword (species, domains, subjets and user defined tags) chip and get a search output for that term.
ATM is a fake dismissable icon.

knirirr:
N.B. there's a small display bug here that I wasn't able to fix: If, for example, one is viewing a route with search?domains=text, then all domain chips will be highlighted and others will not. This looks a bit odd.


hoseinmirian:
@knirirr the behavior of chips should not be additive? I mean after adding each chip to the URL by clicking other chips in the card the previous query will be removed and being replaced by new keyword, is the current behavior  is the one that should happen?

knirirr:
@hoseinmirian I would have thought not, but I don't know - it is a matter for users to answer.

knirirr:
I've created a separate PR here with different behaviour; if that one is preferred then this may be discarded:

https://github.com/FAIRsharing/fairsharing.github.io/pull/431

PR #431, opened 2020-08-27
Keyword links

knirirr:
This pull request replaces https://github.com/FAIRsharing/fairsharing.github.io/pull/422, but I've left both up in case users' preference is for one over another. 

In this PR clicking on links in record cards should add multiple search terms (e.g. both domain and subject), in the other PR only a single term is searched.


terazus (changes requested):
### ISSUE 1:
If you try to apply several times the same filter/value you get the following error in console:
Uncaught (in promise) Error: Avoided redundant navigation to current location: "/search?domains=Mathematical%2520model".
You should first check if the route has the given field and value before trying to apply it.

error

### ISSUE 2:
It's only additive when different fields are selected. If it's additive, then it should also work for different values within the same field:
error2

tests/unit/components/Records/Search/Output/SearchLinkChips.spec.js
```diff
@@ -52,7 +52,22 @@ describe("SearchLinkChips.vue", function () {
         let chip = {label: 'orange', active: 'false'};
         wrapper.vm.updateSearchQuery(chip);
         expect($router.push).toHaveBeenCalledTimes(3);
+
+
     });
 
+    it("allows multiple fields of same tag type", () => {
```
terazus, line 66: This can be improved by adding an extra expect($router.push).toHaveBeenCalledWith({....}) to check the input given to the push mock.

tests/unit/components/Records/Search/Output/SearchLinkChips.spec.js
```diff
@@ -52,7 +52,22 @@ describe("SearchLinkChips.vue", function () {
...
+        wrapper.vm.$route.query = {domains: 'banana'};
+        let chip = {label: 'orange', active: 'false', type: 'domains'};
+        wrapper.vm.updateSearchQuery(chip);
+        expect($router.push).toHaveBeenCalledTimes(4);
```
terazus, line 70: Same as above

terazus (approved):

PR #438, opened 2020-09-07
added pagination to buttom and fixed #406

### LINKED ISSUE #406: Add pagination to the bottom of the search pages
Pagination should appear twice: at the top, and at the bottom;

hoseinmirian:


terazus (approved):

knirirr (approved):

terazus (changes requested):
It create a really strange effect where the skelleton loader is displayed twice which we need to fix:
barbug

knirirr (approved):

terazus (commented):
Im waiting for @knirirr and @ramGranell to comment on what they think but i don't like the solution because you simply deleted the skeleton loader. I'd like to have a skeletton loader on the pagination bar and none for the bottom one.

ramGranell (commented):
For me it is fine but I am not the expert in the topic. There is just a grey box when the page is loading.

hoseinmirian:
> . I'd like to have a skeletton loader on the pagination bar and none for the bottom one

I think it's fine my way. The reason its wrapped with a skeleton loader is because of the keeping consistency in terms of width and height.

terazus (changes requested):
Has a failing unit test + minor styling:
image
It's too close to the last record and if you wanna include the number of items displayed (which I personally wouldn't) then you need to print them **before** the bottom pagination bar.

Issue #405, opened 2020-08-13
Search hits


terazus:
Display the search results hits as "displaying {x} to {y} records out of {total items}" where:
- x is page number * item per page
- y is ((page number + 1) * item per page) -1

PR #441, opened 2020-09-07
Initial implementation for #405.

knirirr:
I've created a very simple component that displays the number of hits and the page, according to the requirements in #405.


terazus (changes requested):
Add some margin to your component so it's not so crumbled with the pagination bar (you can use class="mt-x" where x is a number).
Also, the number calculation is wrong:
image
Here we have 30 to 59 on page 1 where it should be 1 to 29.
So 
`x: function(){return this.currentPage * (this.perPage-1)}`
and
`y: function(){return (this.currentPage * this.perPage) -1}`

terazus:
The formula is still wrong: we get 0 to 29 instead of 1 to 30 on page 1. Add 1 to all results.

package.json
```diff
@@ -25,6 +25,7 @@
     "jquery": "^3.5.1",
     "jsdoc-vuex-plugin": "^1.0.0",
     "lodash": "^4.17.20",
+    "npm": "^6.14.8",
```
terazus: Please delete.

terazus (changes requested):

terazus (approved):

Issue #370, opened 2020-08-04
Record when missing general info fields


terazus:
When no DOI on a record page: either hide the field of show No DOI;
When no abbreviation, remove the empty parenthesis.
same for all the other fields: display either None or hide the field if no data is available.
@terazus would be good to have a picture of the route query to see what record had the problem

PR #445, opened 2020-09-08
Adding the record abbreviation to the search output in both Column/Stack

ramGranell:
I added the abbreviation in both different representations column and stack. About the style of the fonts for the abbreviation can be discussion (waiting curators feedback?). In the record page (e.g. http://localhost:8080/#/2985) the abbreviation is in bold, but I don't understand why as name can be more important. Anyway, the bold in search output is different that the record page. Btw, I also fixed typos in one of the paths in the file that it didn't seem to affect the general behaviour.


terazus (approved):

ramGranell:
Apologies as I found an error: when there is not abbreviation, it still adds the parenthesis, e.g. http://localhost:8080/#/search?q=Bioschemas
It also happened in a Record page (there is already a ticket for this #370 ). 

There is also another  problem when there are long names and abbreviations (e.g. http://localhost:8080/#/search?q=dommino), do we have to do something special about this. Anyway, it seems in this case that abbreviation is not correct (thing to analyse for DB migration). However, in the frontend are we going to do something about this?

Issue #458, opened 2020-09-11
Number of hits should be 0


terazus:
Or shouldn't be displayed when there is no results.

knirirr:
Would you be able to provide an example of this bug? I am not sure what you mean.

PR #461, opened 2020-09-14
Hide hitcounts if hits=0. #458.

knirirr:


src/components/Records/Search/Header/HitCount.vue
```diff
@@ -1,6 +1,10 @@
 <template>
-  <div class="mt-2" style="text-align: center;">
-    <p>Displaying {{x}}  to {{y}} of {{ hits }}.</p>
+  <div
+    v-if="hits"
+    class="mt-2"
+    style="text-align: center;"
```
terazus: Try avoid using style at all costs, especially when there's a vuetify class that does that.
Intead of `style="text-align:center"` use `class="mt-2 text-center"`
(see https://vuetifyjs.com/en/styles/text-and-typography/)

image

terazus (changes requested):

terazus (approved):

Issue #460, opened 2020-09-14
Autocomplete scroll with the body


terazus:
It creates a strange effect where the autocomplete dropdown won't stick to the input field and move when we scroll.

terazus:
Please, if you open a ticket with vuetify, link it here. Or let is know if you need to make contact with the vuetify team or can fix it on your own.

terazus:
Checked quasar implementation, and it's the same thing: https://v0-17.quasar-framework.org/components/autocomplete.html

knirirr:
@hoseinmirian please work on this today and Monday. Opening a ticket as suggested above ASAP would be advisable to give more time to get a response.

PR #468, opened 2020-09-15
Hossein auto complete scrolling bug #460

hoseinmirian:


knirirr (commented):
Is this commit to fix the problem of autocomplete dropdowns staying in position when the screen scrolls? If so then I'm still seeing the bug on my system when I check this branch out.

terazus:
> 
> 
> Is this commit to fix the problem of autocomplete dropdowns staying in position when the screen scrolls? If so then I'm still seeing the bug on my system when I check this branch out.

same here

hoseinmirian:
> 
> 
> Is this commit to fix the problem of autocomplete dropdowns staying in position when the screen scrolls? If so then I'm still seeing the bug on my system when I check this branch out.

yes it is. it should be fine.

terazus (changes requested):
Not working as intended, the behaviour is still here (only the scroll bar has disaper. You now need to scroll locking of the left panel.
After that you to also sort the two position: small height and full height.

knirirr:
The latest change does appear to fix the scroll issue, which is good. However, it has affected the appearance of elements within the dropdown, which now seem to overflow (I'm using Firefox, BTW). Counts with one digit fit, two causes the element to touch the edge of the dropdown, and a third digit causes the overflow.

Screenshot 2020-10-02 at 13 45 41
Screenshot 2020-10-02 at 13 41 57

knirirr (approved):
Please wait for @terazus to review also before merging - thanks.

terazus (changes requested):
Please merge from dev before i do the review :)

src/components/Records/Search/Input/SearchInput.vue
```diff
@@ -107,8 +107,15 @@ export default {
         comparison = 1;
       }
       return comparison;
-    }
-  },
+    },
+/*
+  /please do not remove this function/
```
terazus: Please keep these commented functions either on a local branch or on a separate branch. What you submit ends up on master and unused code shouldn't be merged.

src/components/Records/Search/Input/FilterAutocomplete.vue
```diff
@@ -83,8 +85,7 @@ export default {
         if (_module.selectedValues !== null && _module.selectedValues.length > 0) {
           if (_module.selectedValues.length === 1) {
             currentParams[filterName] = encodeURIComponent(_module.selectedValues.join(','));
-          } else
-            {
+          } else {
```
terazus: Stroustrup on all if/else catch/try plz.

tests/unit/store/uiController.spec.js
```diff
@@ -26,28 +26,36 @@ describe('Mutations & Actions', () => {
...
         actions.commit = jest.fn();
         actions.setScrollStatus(state, fakeState);
         expect(actions.commit).toHaveBeenCalledTimes(1);
+    });
 
+    it("can check the setStickToTop action", () => {
+        let fakeState = true;
+        actions.commit = jest.fn();
+        actions.setStickToTop(state, fakeState);
+        expect(actions.commit).toHaveBeenCalledTimes(1);
```
terazus: use haveBeenCalledWith **if possible** for all these tests.

terazus (changes requested):
Minor modifications and we are good to merge.

terazus (approved):

PR #469, opened 2020-09-15
Dom check can edit

terazus:


knirirr (commented):
This all looks good, though I noticed one small issue:
If I'm looking at a record whilst logged out, then log in using the dropdown, the edit button does not appear - I have to refresh the page. Presumably this is a bit tricky to fix?

PR #485, opened 2020-09-18
Dom 404 page

terazus:
To be tested with some random urls like /ima/random/url or fakes records ids like 123455789.


knirirr (commented):
If I navigate to something like http://localhost:8080/#/banana then I get the expected 404, but omitting the # goes to the home page instead. Is there a means to catch that latter case, as I might expect there to be incorrect URLs (e.g. from old links on the web) without the correct hash in.

knirirr (approved):

PR #490, opened 2020-09-18
Client-side validation for the GeneralInfo edit page

knirirr:
Presently this is merely a draft as it's still a WIP.


knirirr:
I think this is about all I can add at this point, until we have some more edit forms completed.

terazus (approved):

PR #496, opened 2020-09-21
Button to request to maintain a record

knirirr:
This is a draft whilst I deal with two issues relating to this ticket:
1. I need to work out how to do a nice popup modal (suggestions welcome) for when a request is successfully received by the API.
2. This ticket must be completed first: https://github.com/FAIRsharing/FAIRsharing-API/issues/175


terazus:
Netlify builds are likely to fail due to console.log statements. Didn't verify though.

knirirr:
I always check for them. Might have missed one though.

terazus:
In Record.vue:
image

src/components/Client/RESTClient.js
```diff
@@ -269,6 +269,31 @@ class RESTClient {
         return response.data;
     }
 
+    async claimRecord(recordID, userToken) {
```
terazus, line 284: Please add the JSdoc comments for these two methods.
I'm also realizing i forgot to add them for edit and create publications.

src/views/Records/Record.vue
```diff
@@ -16,16 +16,25 @@
...
       >
         <v-spacer />
         <v-btn
+          v-if="canEdit"
           class="success"
           @click="goToEdit()"
         >
           EDIT
         </v-btn>
+        <v-btn
```
terazus, line 30: Do we agree that if a user can both edit and claim, then two buttons would  show up?
You may want to use a v-else-if="mycondition" instead.
Not a request, just a thing to consider.

src/views/Records/Record.vue
```diff
@@ -190,6 +201,31 @@
...
               })
             },
+            async requestOwnership() {
+              let _module = this;
+              const recordID =  _module.currentRecord['fairsharingRecord'].id;
+              const claim = await client.claimRecord(recordID, _module.user().credentials.token);
+              if (claim.error) {
+                _module.error = "Sorry, your request to claim this record failed. Please contact us.";
+                _module.canClaim = false;
+              } else {
```
terazus: can you please put else's on new lines ? :D
The official name is Stroustrup styling.
Here in case you are interested: https://eslint.org/docs/rules/brace-style

My reasoning behind this is that webstorm has issues dealing with collapsing/expanding blocks styled with **one true brace**. It works much better with **Stroustrup**.

src/views/Records/Record.vue
```diff
@@ -190,6 +201,31 @@
...
+              }
+            },
+            async checkClaimStatus() {
+              let _module = this;
+              if (_module.user().isLoggedIn) {
+                const recordID = _module.currentRecord['fairsharingRecord'].id;
+                const claim = await client.canClaim(recordID, _module.user().credentials.token);
+                if (claim.error) {
+                  _module.canClaim = false;
+                } else {
```
terazus: same as above

terazus (changes requested):
A very few minors things to changes:
- [x] missing JSdoc comments 
- [x] Stroustrup styling

src/views/Records/Record.vue
knirirr: This is one I'm not completely sure about. Technically, curators don't need to claim anything, so I was thinking about allowing them to assign maintainer roles to themselves directly in an admin interface to be developed later. But, I also thought they might request that this button show up for them as well.

terazus (approved):

Issue #488, opened 2020-09-18
When viewing a collection, associated records should be reversed


terazus:
or, even best, collects label should be replace by "is collected by" so we don't need to reverse the assocation.
e.g: https://preview.fairsharing.org/#/3118

Extands to policies recommanding a record. Label should be "is recommended by"
e.g.: https://preview.fairsharing.org/#/FAIRsharing.dIDAzV

PR #499, opened 2020-09-22
Rg associated records label reversed table #488

ramGranell:


tests/unit/views/Records/Record.spec.js
```diff
@@ -152,7 +152,25 @@ describe("Record.vue", function() {
                 "recordAssocLabel": "collects"
             },
         ];
-        wrapper.vm.prepareAssociations(fakeAssociatedRecords,fakeReverseAssociatedRecords)
+        wrapper.vm.prepareAssociations(fakeAssociatedRecords,fakeReverseAssociatedRecords);
```
terazus: There are missing expect statements in this test.

terazus (changes requested):
Improve unit tests

terazus (approved):

Issue #495, opened 2020-09-21
Additional buttons for edit form


knirirr:
We will need:
- An "exit" button to return to the record page without saving anything.
- A "clear changes" button which will reload the existing content from the store. 
The latter will be needed on each sub-form but the former could go separately, perhaps at the top.

Issue #373, opened 2020-08-04
Editor submit button


terazus:
Change submit button of general info and keywords to Submit {{section}}

PR #515, opened 2020-09-24
Edit form buttons

knirirr:
Some modifications to the edit form based on tickets #373 and #495.
The latter ticket should allow reloading data from the database (to overwrite changes) and also exiting from editing without saving.


knirirr:
Having looked into this I think that some further discussion is needed.
However, the contents of this PR could well be useful as they are (there was a specific request for the "exit editing" button on the old version) and so I'll request review with a view to a perhaps making new ticket for any further enhancements.

src/views/CreateRecord/Editor.vue
```diff
@@ -3,6 +3,45 @@
     id="recordEditor"
     fluid
   >
+    <!-- TODO: Loop through the buttons -->
+    <!-- popup to confirm exit from editing -->
+    <v-list>
+      <v-list-item
```
terazus: is wrappring the dialog boxes inside a list really usefull ? i would have wrapper it into a `<v-row> <v-col>` instead

terazus (approved):

src/views/CreateRecord/Editor.vue
knirirr: I have no idea which is better. A list seemed like the appropriate choice to me (seems to be used elsewhere in the code). If you want me to change it let me now, otherwise I'll leave it as it is.

Issue #426, opened 2020-08-19
after clicking on any button, scrollToTop method must be called. 


hoseinmirian:

ramGranell:
Hi, I helped @hoseinmirian in a pair coding session when it was the power failure. We managed to repair it partially but not totally. I haven't done anything more and it seems that it is not completely repaired.

hoseinmirian:
@knirirr @terazus this issue is useless anymore due to there is a pr for that now

PR #537, opened 2020-10-05
Hossein scroll to top after button click #426

hoseinmirian:


terazus (changes requested):
This PR needs a lot of review and I think there's a deeper problem of implementation of the layout that creates the need to hack stuff around.

Issues (related to that component in general): 

- [x] duplication of the scrollutils file with the goto top() method.
- [x] wrapping an export inside a method, itself wrapped in another method inside setTimeout of a watcher. All of that to scroll to the top? This should be an indicator that something is going wrong.
- [x]  duplication of a watcher/computed property.
- [x] minor styling issues (whitespace after colon + single line when DOM elements only have one property).
- [x] could simplify the map actions imports to avoid duplication (2 wrappers around store action).
- [x] writing and completing tests.

very hard to modify/maintain/update any piece of code related to layout/navigation/overflow on the search page. There's a small probability that we need to review the whole implementation of the overflow in order to have a maintainable code.

knirirr:
@hoseinmirian the tests are not passing (see: https://github.com/FAIRsharing/fairsharing.github.io/runs/1241775562). This will need to be rectified before this PR can be reviewed.

terazus (changes requested):
Did you run the tests locally ? What does the output tells you ?
There are 2 unhandled promise rejections (+ an extra one that will NOT show up locally and you can ignore).
image

Issue #326, opened 2020-07-22
Choose Material Design icons as the final icon source of the project


hoseinmirian:
All Font Awesome icons must be replaced with Material Design icons.

knirirr:
In addition, some additional icons may need to be found, such as Pubmed, DOI, title and something for organisations other than a factory. See: https://docs.google.com/document/d/1g1sp9vSFEfkLw6pyiESmjdxRRUiN9tjiz8k7Gbv7_VE/edit# (section 3).

PR #539, opened 2020-10-08
Hossein replacing all icons with md iicons #326

hoseinmirian:


terazus (changes requested):
Still a lot of font awesome icons remaining:
image

Issue #538, opened 2020-10-07
Redirect user after login in


terazus:
When there is no redirect instruction, the user should be redirect to his profile page when loggin in.

PR #540, opened 2020-10-08
Dom more login redirect

terazus:
Fixing #534 and #538


knirirr:
I tried this out and noticed a couple of things - I'm not sure if these are known.
1. After I confirmed my email address and logged in I didn't get redirected to my profile page.
2. When I tried to claim a record I got: click "claim" -> log in -> redirected back to record page -> click "claim" again -> claim banner displayed.
Is this expected behaviour?

terazus:
Answering your comments:
1. You were probably redirected to master when using the link from the email, where the modifications don't exists yet :)
2. So far yes. If we want to make it more "automatic" we need a dedicated view for the claim record feature (like /:id/claim) that handles the claim and redirects back to the record page with either an error or a confirmation message. It's a bit harder to do though.

PR #549, opened 2020-10-14
added all missing icons from alerts: fixed #548

### LINKED ISSUE #548: Still missing icons
There are still some fontawesomes icons missing, especially in all alerts:
image

terazus:


hoseinmirian (changes requested):
@terazus 
I checked why font-awesome was not loading properly, found out with only moving those lines from where it is to icon.js, the problem will be solved and we will not need anymore like now override the icons.. all the default value become FA. even I checked with the pagination I checked remove the prev and next icon of it and it did work.. 
image

hoseinmirian (approved):

PR #554, opened 2020-10-21
trying to fix unit tests bugs en record.spec.js

terazus:


knirirr (approved):
I didn't see any more errors when testing this on a Linux system so I presume that the various additional mocks were effective.

PR #556, opened 2020-10-23
Test failure seems to be related to line 327 in Record.vue.

knirirr:
As per the title.
I presume that the tests are set up in such a way that the response to checkClaimStatus is being checked but doesn't contain 'data'. I may be wrong, though.


knirirr:
This has been dealt with by https://github.com/FAIRsharing/fairsharing.github.io/pull/557

PR #565, opened 2020-11-10
added reusable tooltips for DRAO/SRAO terms

terazus:


terazus:
tooltips

terazus:
tooltips

knirirr (approved):
This looks great, although I'm certain that curators won't want to see the expanded names in the tooltip - this field is only present as a hack to simulate the effects of traversing the ontology within a text search.

PR #568, opened 2020-11-13
Dom code splitting

terazus:


knirirr (commented):
The code seems fine and runs on my laptop (Mac still broken). However, I've noticed a drop in coverage on both systems, in src/router/index.js.
Do you also see that?

knirirr (approved):

PR #592, opened 2020-12-02
Hossein fix fiber node update #591

### LINKED ISSUE #591: Fiber issue with new node js
there is a Fiber package in the package.json that was not working with new node js version. So due to I have the latest node js version I was in trouble. needed to update this package.

hoseinmirian:


package.json
```diff
@@ -22,6 +22,7 @@
     "core-js": "^3.6.5",
     "deep-object-diff": "^1.1.0",
     "dotenv": "^8.2.0",
+    "fibers": "^5.0.0",
```
terazus: This is not a satisfying solution because fibers cannot be installed without a python environement. It was in optionalDependencies for the reason specified above:

image
From https://docs.npmjs.com/cli/v6/configuring-npm/package-json#optionaldependencies

I would try to bump the version to 5.x in the optional dependency and see if it works.

terazus (changes requested):

ramGranell (approved):

terazus (approved):

knirirr (approved):

PR #600, opened 2020-12-03
added Footer in Records in a full-width style

hoseinmirian:


hoseinmirian:
please note that this will put Footer below all content to be able showing Footer full-width. Ofcourse its not showing Footer fully due to upcoming branchs will fix this issue.

knirirr (approved):
Presumably this is related to the previous PR on which I commented; it can be hard to work out what's happening unless there's a comment explaining it, other than the title.

src/views/Records/Records.vue
```diff
@@ -72,10 +71,11 @@ import JumpToTop from "@/components/Navigation/jumpToTop";
 import recordsLabels from "@/data/recordsTypes.json"
 import FilterChips from "@/components/Records/Search/Header/FilterChips";
 import filterChipsUtils from "@/utils/filterChipsUtils";
+import Footer from "@/components/Navigation/Footer";
 
 export default {
   name: "Records",
-  components: { JumpToTop, SearchOutput, SearchInput, FilterChips},
+  components: {Footer, JumpToTop, SearchOutput, SearchInput, FilterChips},
```
terazus: Why not handle the footer from App.vue by removing the v-if in the footer?

terazus (changes requested):

src/views/Records/Records.vue
hoseinmirian: You are right we can do it. its been done in the next branches. so this will be overwritten..

PR #609, opened 2020-12-04
Hack to correct coverage locally.

knirirr:
There's been an irritating issue for a while where coverage is shown correctly as 100% online, but running tests locally show somewhere between 95-96%.
A possible hack to "fix" this is to exclude the offending lines from coverage. Unfortunately, the means of doing this is a bit ugly as it's not a single function that's causing it but instead a list of statements, each one of which must be commented.
I've left this as a draft PR as it's not necessarily something that's worth doing.
Please feel free to try it and comment, when time permits.

### DIFF
--- a/src/router/index.js
+++ b/src/router/index.js
@@ -2,34 +2,58 @@ import Vue from "vue";
 import VueRouter from "vue-router";
 import store from '@/store'
 
-
+/* istanbul ignore next */
 const Home = () => import(/* webpackChunkName: "home-chunk" */ '@/views/Home/Home.vue');
+/* istanbul ignore next */
 const NotFound = () => import(/* webpackChunkName: "home-chunk" */  "@/views/Errors/404");
 
+/* istanbul ignore next */
 const Record =  () => import(/* webpackChunkName: "record-chunk" */ '@/views/Records/Record');
+/* istanbul ignore next */
 const Records =  () => import(/* webpackChunkName: "records-chunk" */ '@/views/Records/Records');
+/* istanbul ignore next */
 const NewRecord = () => import(/* webpackChunkName: "newRecord-chunk" */ '@/views/CreateRecord/NewRecord');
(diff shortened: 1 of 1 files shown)


terazus:
To make it a bit cleaner you can declare all the const as null or as nothing (e.g.: const Home; or const Home = null).
Then you declare a function that initiates all the dynaamic imports and adds the "ignore comment" to the whole function itself.

knirirr:
Are you able to offer an example of what you have in mind? I'd interpret that as something like:

```javascript
const Home;

function dynamicImports() {
    Home = () => import(/* webpackChunkName: "home-chunk" */ '@/views/Home/Home.vue');
}

dynamicImports();
```

But `Const declarations require an initialization value (6:10)`. Assigning the initial const as null doesn't allow reassignment.

terazus:
You are on the right track. The issue is that **const** can't be modified. That's why they are **const**.
Use a **let** instead and it should work :)

knirirr:
Indeed, though I was not certain that it would be OK to not have these as constants.
I've pushed a variant where I freeze each after assignment; still verbose, but not quite as ugly I think.

src/router/index.js
```diff
@@ -2,35 +2,92 @@ import Vue from "vue";
...
+let OauthLogin;
+let LoginFailure;
+let Stat;
+let Community;
+let Stakeholders;
+let Timeline;
+let License;
+let Terms;
+let Educational;
+let Privacy;
```
terazus: You can declare the variables with commas without having to reuse the LET keyword. Works on single and multiple lines.
```suggestion
let License, Terms, Educational, Privacy, ... ;
```

terazus (changes requested):

knirirr:
I'm less keen on the look of it, but not particularly concerned so I've changed it.

terazus (approved):

Issue #584, opened 2020-12-01
Countries should be dissmissible


terazus:
In the edit form, general section, countries should be removable using the mouse.

PR #611, opened 2020-12-04
Countries can be dismissed by click. #584.

knirirr:
This works. But, when a country is dismissed by click, the drawer pops open. I'm not keen on this but, unfortunately, it appears to be a bug in Vue.js that adding styling to an autocomplete produces this behaviour. Example:
https://stackoverflow.com/questions/60577675/how-to-prevent-v-autocomplete-from-opening-the-dropdown-drawer-when-clicking-on
If no-one objects to the behaviour here then it might suffice. 
Of course, someone who is actually a Javascript programmer may think of a means to fix this. ;-)

### DIFF
--- a/src/components/Editor/EditGeneralInfo.vue
+++ b/src/components/Editor/EditGeneralInfo.vue
@@ -142,8 +142,8 @@
                 :hint="descriptions['countries']"
                 item-text="name"
                 item-value="name"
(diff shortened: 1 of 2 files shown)


knirirr:
I'll merge this as discussed, but open a new ticket

PR #613, opened 2020-12-04
Rg subs doi table for download button#590

ramGranell:
Not sure about the test of the function to check the file creation and saved


knirirr (approved):

src/views/Curators/Curator.vue
```diff
@@ -510,6 +464,16 @@
...
+          },
+          async obtainFileRecordsWODois(){
+            let data = await restClient.getRecordsWoDOIs(this.user().credentials.token);
+            let dataStr = "data:text/json;charset=utf-8," + encodeURIComponent(JSON.stringify(data).replace(/^\[(.+)\]$/,'$1').split(',').join('\r\n').replace(/['"]+/g, ''));
+            let downloadAnchorNode = document.createElement('a');
+            downloadAnchorNode.setAttribute("href",     dataStr);
+            downloadAnchorNode.setAttribute("download", "recordsWithoutDOIs.txt");
+            document.body.appendChild(downloadAnchorNode); // required for firefox
+            downloadAnchorNode.click();
+            downloadAnchorNode.remove();
```
terazus: You don't need to do this.
You can wrap your icon inside an `<a>` tag and return the href encoded variable.
You would then have:
`<a :href='obtainFileRecordsWODois()' download="recordWithoutDOIs.txt"> <v-icon> fa-download </v-icon></a>`
and then

```suggestion
           return dataStr
```

terazus (changes requested):

src/views/Curators/Curator.vue
ramGranell: I try the suggested changes but two things happen:
1) The file is not properly created and it appears this text in the downloaded file:
<!DOCTYPE html>
<html lang="en">
<head>
    <meta charset="utf-8">
    <meta http-equiv="X-UA-Compatible" content="IE=edge">
    <meta name="viewport" content="width=device-width,initial-scale=1.0">
    <link rel="icon" href="/favicon.ico">
    <title>Webpack App</title>
    <link rel="stylesheet" href="https://fonts.googleapis.com/css?family=Roboto:100,300,400,500,700,900">
<link href="/js/app.js" rel="preload" as="script"><link href="/js/chunk-vendors.js" rel="preload" as="script"></head>
<body>
<noscript>
    <strong>We're sorry but Webpack App doesn't work properly without JavaScript enabled. Please enable it to continue.</strong>
</noscript>
<div id="app"></div>

<script type="text/javascript" src="/js/chunk-vendors.js"></script><script type="text/javascript" src="/js/app.js"></script></body>
</html>

2) The nice button with the text "obtain file" is substituted just with the icon with the hyperlink that does not have the same quality and looks worse than before.

Can I keep the old implementation that works properly and looks better?

src/views/Curators/Curator.vue
```diff
@@ -258,7 +257,6 @@
           maintenanceRequests: [],
           recordsCreatedCuratorsLastWeek: [],
           recordsInCuration: [],
-          recordsWithoutDois: [],
           hiddenRecords: [],
           recordType: null,
           headers: {
```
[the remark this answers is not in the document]
ramGranell: In fact, I don't need to have recordsWithoutDois longer and they can be removed.

src/views/Curators/Curator.vue
ramGranell: @terazus I uploaded the suggested changes and a file is downloaded but it is not properly obtained. I think is related that it tries to obtain it from localhost instead that the call to the restClient is correct.

terazus (approved):

PR #626, opened 2020-12-11
Rg curator pending request table#615

ramGranell:


terazus (dismissed):

knirirr (changes requested):
One minor issue is that one of the icons renders the wrong size:

Screenshot 2021-01-18 at 13 27 21

One other thing is this dialog box:

Screenshot 2021-01-18 at 13 33 50

It would be useful to disable each of the buttons when it's clicked; at present it can be clicked multiple times and will send multiple requests to the server. In this case it doesn't cause any serious problem but it will result in the record being saved twice if a user double-clicks this button, for example.

terazus:
@ramGranell an easy way to fix the icons size is to use the vuetify avatar component: https://vuetifyjs.com/en/components/avatars/
An easy way to fix the second problem depends on what component you use but it's very likely just adding a `persistant` tag to the component

ramGranell:
> @ramGranell an easy way to fix the icons size is to use the vuetify avatar component: https://vuetifyjs.com/en/components/avatars/
> An easy way to fix the second problem depends on what component you use but it's very likely just adding a `persistant` tag to the component

I did both suggestions. However some of the icons seem to have different size in the image (I put a fix image size). Anyway, I know that images need to be changed and standardised to have same size, aren't they?

knirirr:
I tried this branch out at https://deploy-preview-626--fairsharing.netlify.app/ and was still able to double-click the button to generate multiple requests - take a look in the network tab and you should be able to reproduce it.

ramGranell:
> I tried this branch out at https://deploy-preview-626--fairsharing.netlify.app/ and was still able to double-click the button to generate multiple requests - take a look in the network tab and you should be able to reproduce it.

I modified the code to have a variable to disallow all the buttons once any button is pressed. All buttons are allowed again when any menu with the buttons is opened again.

ramGranell:
Just to remember that the table to review in this PR is just the table with title "OWNERSHIP REQUESTS ". Once it is approved, I will use some of the accepted modifications in this table (e.g. icon size, button disallowed after clicking it) to the top table.

knirirr (approved):

src/views/Curators/Curator.vue
```diff
@@ -412,17 +310,26 @@
...
               let object = {};
               object.createdAt = item.createdAt;
-              object.recordNameID = item.fairsharingRecord.name+' ('+item.fairsharingRecord.id+')';
-              object.userNameID = item.user.username+' ('+item.user.id+')';
+              object.recordName = item.fairsharingRecord.name + ' ('+ item.fairsharingRecord.id +')';
+              object.id = item.fairsharingRecord.id;
+              object.type = item.fairsharingRecord.type ;
+              object.userNameID = item.user.username + ' ('+item.user.id+')';
+              object.processingNotes = item.fairsharingRecord.processingNotes;
+              object.requestID = item.id;
```
terazus: Single object with string interpolation:
```suggestion
              let object = {
                createdAt: item.createdAt,
                recordName: `${item.fairsharingRecord.name} (${item.fairsharingRecord.id)})`,
                id: item.fairsharingRecord.id,
                type: item.fairsharingRecord.type,
                userNameID: `${item.user.username} (${item.user.id})`,
                processingNotes: item.fairsharingRecord.processingNotes,
                requestID: item.id
            };
```

terazus (changes requested):

src/views/Curators/Curator.vue
ramGranell: In the last commit, I changed manually all the object declaration to init the fields there as comment suggested.

terazus (approved):

PR #631, opened 2020-12-15
Dom new editor

terazus:


knirirr:
Is this function used for anything in particular, or planned to be? It seems only to be referenced here:
https://github.com/FAIRsharing/fairsharing.github.io/blob/master/src/store/record.js#L57-L63

knirirr (changes requested):
I've opened some tickets for issues that can be done after this is merged. Before merging there's only one more thing I think needs to be fixed, which is that it's possible to add duplicate tags:

Screenshot 2021-01-06 at 11 05 11

Apparently, that's here: https://github.com/FAIRsharing/fairsharing.github.io/blob/DomNewEditor/src/components/Editor/GeneralInformation/NewTags.vue#L184-L202

terazus:
> 
> 
> Is this function used for anything in particular, or planned to be? It seems only to be referenced here:
> https://github.com/FAIRsharing/fairsharing.github.io/blob/master/src/store/record.js#L57-L63

This is an artifact from the old version. If I remember correctly i still have another tab that relies on this action so untill everything is correctly refactored, we still need it

knirirr (approved):

hoseinmirian (changes requested):
1- I noted that action is not used and changed value directly from the mutations.. please explain if it's been intentional and if yes why. 
image
image


2- Webstorm warns about an invalid number of arguments:
image

3- You missed an await here
image

4- better to not have inline style unless it's a dynamic styling
image

5- extra style tag in this component
image

hoseinmirian (approved):

PR #634, opened 2020-12-17
Fix header absolute position

hoseinmirian:
it is a quick fix on header bug with absolute position. 
also removed claim phrase from add/claim content..


terazus:
Can"t pull this branch:
image

terazus (changes requested):
please ignore this message as im unable to delete it because github it stupid

terazus (approved):

PR #643, opened 2021-01-07
Detect if users attempt to create a dataset

knirirr:
Once this is merged then I can also merge the relevant API code:
https://github.com/FAIRsharing/FAIRsharing-API/pull/219#pullrequestreview-563427689

What this does is set a flag ("is_dataset") which is passed to the API to deal with, if the following user behaviour is encountered:

1. Selecting a "database" of some sort as the type of their new record.
2. Clicking "yes" when asked if they intend to upload a dataset.
3. Clicking "go back" when seeing the message that says they should stop because we don't accept datasets.
4. Selecting "database" again.
5. Clicking "no" when the question comes up again.

At present it will be triggered any time a database record is selected rather than just on create. But, disabling the changing of types other than for curators needs to be done.


ramGranell (approved):

src/components/Editor/GeneralInformation/DatabaseWarning.vue
```diff
@@ -113,6 +113,13 @@
           pressYes(){
             this.answered = true;
             this.submitted += 1;
+          },
+          pressNo(){
+            this.showOverlay = false;
+            if (this.submitted === 1) {
```
terazus:
```suggestion
            if (this.submitted >= 1) {
```
=> so if a user presses more than once, it also sends the email :)

terazus (changes requested):

terazus (approved):

Issue #638, opened 2021-01-05
Changing registry/type - button disabling


knirirr:
This ticket https://github.com/FAIRsharing/FAIRsharing-API/issues/215 makes reference to schema validation errors which are (correctly) thrown when a record which contains data not permitted in other record types is changed to another type.

As discussed there, modifying the metadata and warning the user would be complex. As this is a feature which is not likely to be used often this complex functionality could be left until after MVP. To prevent users being confused by the errors mentioned above, the type/registry field could be made read-only on editing unless the user is a curator; curators will know what they've done and understand why they must change the metadata, and how.

PR #644, opened 2021-01-07
Limit changing of record type. #638

knirirr:
This change disables the record type/registry field unless:

1. Creating a new record, or
2. Logged in as a curator.

The curators will want users to always be able to change registry/type later but that will involve more complex operations with the stored data and so is best left until after MVP (users can't change types in the current system anyway). 

This can be closed afterwards: https://github.com/FAIRsharing/FAIRsharing-API/issues/215


src/components/Editor/GeneralInformation/BaseFields.vue
```diff
@@ -364,16 +365,25 @@
...
+                  obj.label !== country.name && obj.id !== country.id
+              );
+          },
+          typeChangeDisabled(){
+            let _module = this;
+            if (_module.getCreatingNewRecord) {
+              return false;
+            } else {
+              return !_module.user().is_curator;
             }
```
terazus: else is not required:
```suggestion
            if (_module.getCreatingNewRecord) {
              return false;
            } 
            return !_module.user().is_curator;
```

src/store/record.js
```diff
@@ -94,6 +95,9 @@ let recordStore = {
         resetRegistry(state){
             state.sections.generalInformation.data.type = "";
         },
+        setCreatingNewRecord(state){
```
terazus, line 98: Don't you also need a method to set it back to false when users are editing a record after creating a new record ?

terazus (changes requested):

src/store/record.js
knirirr: Good point, I'll have to look into that.

hoseinmirian (approved):

terazus (approved):

PR #650, opened 2021-01-14
Editing publications

knirirr:
Using the store appears to work. But, all tests are broken in a way that doesn't immediately seem related to the modifications.


knirirr:
Some other issues that need attention, other than the failing tests:

- [x] Changes made are not correctly counted by deep-object-diff.
- [x] The isCitation value of a record is not correctly handled by the store; probably related to it being a property in metadata.

knirirr:
Please note when reviewing that what I've done for checking the number of changes is to create an object which contains only the IDs of the publications and whether they're citations (this latter is set on mounting the component). 
Then, for the comparison, I can count the number of changes of isCitation plus the number of objects not in the union of original and current (line 384).

ramGranell:
It is great, just a couple of comments:
1) How do you create a new publication for the record or is it a different ticket?
2) The column headers of the search table need to be starting with uppercase and "Pubmed ID" appears located a little higher than the other column header. By the way, the order of the headers is also a little stange, the DOI is in the second column and Pubmed ID (only medicine related papers have this). I think authors and journal are more important and informative for an article

knirirr:
@ramGranell re point 1, if your search fails to turn up what's required a button will appear inviting the creation of a new one:

Screenshot 2021-01-22 at 09 48 59

This is the code written by @terazus , which I've not modified - give it a try.

For no. 2, I don't see the height difference for pubmed ID myself. Capitalising the column should be no problem. AFAIK the columns are what the curators have requested (which could therefore change at any time etc. etc.).

ramGranell (approved):

src/components/Editor/EditPublications.vue
```diff
@@ -1,4 +1,4 @@
-<template xmlns:v-slot="http://www.w3.org/1999/XSL/Transform">
+<template xmlns:v-sl ot="http://www.w3.org/1999/XSL/Transform">
```
hoseinmirian: it should be a v-slot. please revert it back to what it was.

src/components/Editor/EditPublications.vue
```diff
@@ -329,24 +335,79 @@
...
+        watch: {
+          publications: {
+            deep: true,
+            handler(newVal){
+              let _module = this;
+              /* istanbul ignore next */
+              if (!_module.initialized) return;
+
+              let updated = _module.checkChanges(newVal);
+              /*
```
hoseinmirian: please use double * so the document generator understands to compile it. now it's known as comment. in case you were intending to actually comment, please ignore this. 

/**
* comment
*/

src/components/Editor/EditPublications.vue
```diff
@@ -62,7 +58,8 @@
                 <div> <b>PUBLISHED IN:</b> {{ publication.journal }}, {{ publication.year }}</div>
                 <div> <b>AUTHORS:</b> {{ publication.authors }}</div>
                 <v-switch
-                  v-model="publication.isCitation"
+                  :input-value="publication.isCitation"
+                  @change="toggleCitation(pubIndex)"
```
hoseinmirian: here eslint suggest to order directives this way:

                  :input-value="publication.isCitation"
                  color="green"
                  label="Cite record using this publication?"
                  @change="toggleCitation(pubIndex)"

image

src/components/Editor/EditPublications.vue
```diff
@@ -166,10 +163,18 @@
...
+          class="primary"
+          :loading="loading"
+          @click="saveRecord(false)"
+      >
+        Save and continue
+      </v-btn>
+      <v-btn
+          :loading="loading"
+          class="primary"
+          @click="saveRecord(true)"
```
hoseinmirian: here we have another eslint warning. 
image

src/components/Editor/EditPublications.vue
knirirr: This typo has been fixed.

src/components/Editor/EditPublications.vue
knirirr: This is merely a comment and is not intended for the documentation.

hoseinmirian (approved):

terazus (approved):

Issue #624, opened 2020-12-11
implementing new stack/grid search card 


hoseinmirian:
- [x] setting up new color palette
- [x] Designing Grid
- [x] writing Grid tests
- [x] Designing Stack
- [x] writing Stack tests
- [x] writing class diagram

PR #651, opened 2021-01-14
Hossein new stack grid card #624

hoseinmirian:


knirirr (changes requested):
I've noticed some things that will need attention.
1. A small issue, but the text "one more tags" should read "one more tag".
2. There's a problem with long descriptions overflowing.
3. The header has changed - compare with current dev., for example.

Images for 2 & 3 are attached (the image for 3 also shows 1)
Screenshot 2021-01-15 at 11 41 15
Screenshot 2021-01-12 at 11 20 22

.

hoseinmirian:
> A small issue, but the text "one more tags" should read "one more tag".

Thanks for your comments. 
1- I will change it. 

2 & 3 I suspect your branch is not having the last commit:

No 2 is committed and should have been fixed already by this commit :
https://github.com/FAIRsharing/fairsharing.github.io/pull/651/commits/ec9161ff10fa03e7fdc156af19f6f9c4e2cab98d
please check the css class name text-ellipses-height-2lines is added in this commit and you should have it by your last commit

image


No 3 is committed: 
 "Revert Header to dev branch"   
https://github.com/FAIRsharing/fairsharing.github.io/pull/651/commits/18f58c50d9f30195b83c77b7db388a7c99966680

Please make sure your branch is updated and see if the problems persist.

knirirr:
I pulled this branch not long before making my comments; there's been one small change since then.
Anyway, the latest pull does indeed seem to have addressed the matters I brought up.

knirirr (approved):

src/components/Records/Shared/AbbreviationTooltip.vue
```diff
@@ -0,0 +1,46 @@
+<template>
+  <v-tooltip
+    v-if="abbreviation!==null && abbreviation.length > 40"
```
terazus, line 3: Do you have an example to showcase this feature please?

src/utils/recordsCardUtils.js
```diff
@@ -1,28 +1,85 @@
...
             } else {
                 return record.id;
             }
+        },
+        getChipColor(chip){
+            if(chip.type==='subjects'){
+                return 'subject_color';
+            }
+            else if(chip.type==='domains')
+            {
```
terazus: why do you put the { on the same line at line 12 and then go to new line at 16 and 20 ?

src/components/Records/Shared/KeywordTooltip.vue
```diff
@@ -7,12 +7,16 @@
...
         <div
           v-bind="attrs"
+          class="ellipse-width-80 text-center"
           v-on="on"
         >
           {{ keyword.label }}
         </div>
       </template>
       <div class="tooltip">
+        <div :class="getChipColor(keyword)">
```
terazus, line 17: you need to add a condition to make sure the type is present or you need to modify the edit general information panel to also pass the type.
Otherwise, we get something weird:
image

terazus (changes requested):

src/components/Navigation/Header.vue
```diff
@@ -1,6 +1,5 @@
 <template xmlns:v-slot="http://www.w3.org/1999/XSL/Transform">
   <v-app-bar
-    id="mainHeader"
```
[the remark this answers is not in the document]
hoseinmirian: no pb. I return it back. thought its not useful!

src/components/Records/Shared/AbbreviationTooltip.vue
hoseinmirian: yes everytime there is an abbreviation available so its not null and its length its bigger than 40 then we ll show the tooltip. see the below picture. 

image

src/utils/recordsCardUtils.js
hoseinmirian: thanks. it will be fixed

src/components/Records/Shared/KeywordTooltip.vue
hoseinmirian: thanks for spotting this unpredicted bug. I ll check that

src/utils/recordsCardUtils.js
```diff
@@ -1,28 +1,83 @@
...
+        getChipColor(chip){
+            if(chip.type==='subjects'){
+                return 'subject_color';
+            }
+            else if(chip.type==='domains') {
+                return 'domain_color';
+            }
+            else if(chip.type==='taxonomies') {
+                return 'taxonomic_color';
+            }
```
terazus, line 20: You need to harmonize the editor colors with these.
You can find them in store/editor.js -> state.colors

terazus (changes requested):
We are missing the "recommended" labels (the red banner at the top right)
Did we decide to remove them ? Or to replace them (with what) ?
E.G: https://deploy-preview-651--fairsharing.netlify.app/#/search?fairsharingRegistry=Standard&isRecommended=true&status=deprecated How do we know this record is recommended?

src/utils/recordsCardUtils.js
hoseinmirian: tnx.recommended is removed based on the latest approved mocks(the decision was made) but there are chances yet in the future to use them but not in a ribbon way. about the editor color I check that.

terazus (approved):

Issue #664, opened 2021-01-27
Record last edit date


knirirr:
A 'lastedited' field has been added to the available data, as per this ticket: https://github.com/FAIRsharing/FAIRsharing-API/issues/232
This field needs to be added to the client so that @hoseinmirian can use it for the record display page.

PR #665, opened 2021-01-27
Added two record query fields (lastEdit, lastEditor). #664.

knirirr:
This adds two fields to the query used for the record display page; the last user to edit a record and the time it was last edited. These could be null.
The criteria for an edit is that an 'update' operation is performed by a user, i.e. acting via a REST controller, and that something in the record changes. The server will determine this by looking through the PaperTrail records, so if requirements are changed the modification will be server-side.


knirirr:
This commit: https://github.com/FAIRsharing/fairsharing.github.io/pull/665/commits/123f9f6ac5d5d73d9f92fb8f1134423088bf54bc
Fixes a broken test for maintenance_requests in the curator dashboard, apparently caused by some functions being deleted from record.js.

terazus (approved):

PR #673, opened 2021-01-29
Scrolls to top after being back from record to records

hoseinmirian:
by moving the line scrollTop as the first command after records page is destroyed as well as minor change in the y axis of the window.scroll.top.y in record destroyed method, I fixed the problem(not scrolling to top by default) when we press back from record display to records. Now whenever and wherever the scroll is in record display when we back to previous page records, we can see scroll is on top..

bug shot after being back from record to records:
106295747-b8a63600-6248-11eb-8e9d-3bd87eb2fd90


knirirr (approved):

src/views/Records/Records.vue
```diff
@@ -122,13 +122,13 @@ export default {
     });
   },
   destroyed() {
+    window.scrollTo(0, 0);
```
terazus, line 125: why are we still using window.scrollTo instead of vue-scrollTo ?

terazus (approved):

src/views/Records/Records.vue
hoseinmirian: due to I need the window.top.scrollY for some calculation in records but v-scrollTo is able to give body specifications not window object. I need window object in some cases but I will investigate to see if I can unify window scrolls into vue-scrollTo..

PR #675, opened 2021-02-01
Hossein redesign general info

hoseinmirian:


knirirr (changes requested):
I like the restructuring of this.
Some things did pop out at me, though:
1. There's a change in the CSS which causes lines to appear under highlighted links pretty much everywhere on the search page and for buttons on the top bar. This will need to be fixed.
2. Is there much point in using some sort of strftime equvalent for getCurrentDate() in Citations.vue, e.g. https://github.com/samsonjs/strftime ? I was surprised to find that Javascript doesn't have that already.

hoseinmirian:
1- thanks for spotting the issue. I will fix it
2- I will use a very famous one https://www.npmjs.com/package/vue-moment for the showing a more human-freindly time and date.

src/components/Records/Record/GeneralInfo/Citations.vue
```diff
@@ -28,7 +28,7 @@
             </span>
             <b v-if="getField('doi')">DOI:</b>{{ getField('doi') }};
             <b v-if="getField('lastEdited')">Last Edited:</b>{{ getField('lastEdited') }}
-            <b>Last Accessed:</b>{{ getCurrentDate() }}
+            <b>Last Accessed:</b>{{ new Date() | moment("dddd, MMMM Do YYYY") }}
```
knirirr: I note that the time is not displayed here; in the old system it is, and so ought to be added for consistency.

knirirr (changes requested):

src/components/Records/Record/GeneralInfo/Citations.vue
hoseinmirian: okay it will be added.

hoseinmirian: image
done

knirirr: Thanks. I'd prefer to see that in 24-hour time as 15:08 (we probably don't need the seconds either), otherwise I think that should do it.

hoseinmirian: okay I remove second and also turn It into a 24-hour time

hoseinmirian: @knirirr 
your request is done. 24 hour time in new commit. 

image

knirirr: We inch closer... just the "am/pm" to dispense with now.

hoseinmirian: image

done.

knirirr: Please remove the "now" then it should be ready to go.

hoseinmirian: image
@knirirr please check

hoseinmirian: image
deleted now.

knirirr (approved):

PR #685, opened 2021-02-03
Dom edit data access

terazus:
Contains:
- the edit data access pannel
- a mechanisms to warn the users when they leave the edit page while there are unsaved changes.
- a fix for dates in curator's unit tests.


src/views/CreateRecord/Editor.vue
```diff
@@ -52,18 +52,24 @@
...
+            class="default"
             @click="panelData.show = true"
           >
             {{ panelData.name }}
           </v-btn>
+          <v-btn
+            class="ml-2 default"
+            :href="'#/' + $route.params.id"
+          >
+            Exist editing
```
knirirr: A typo has crept in here (exit -> exist).

knirirr (changes requested):
I encountered a problem with the confirmation dialog when exiting editing; the first time it was triggered all worked as expected, but on subsequent occasions the exit editing button did nothing. 
I couldn't spot any other issues though.

src/views/CreateRecord/Editor.vue
terazus: It's an issue with the link itself, not the code (try existing clicking on standard for instance, it works flawlessly).

terazus:
It's an issue with the link itself, not the code (try existing clicking on standard for instance, it works flawlessly).

knirirr:
> It's an issue with the link itself, not the code (try existing clicking on standard for instance, it works flawlessly).

Definitely not working for me, I can reproduce the problem.

terazus:
> 
> 
> > It's an issue with the link itself, not the code (try existing clicking on standard for instance, it works flawlessly).
> 
> Definitely not working for me, I can reproduce the problem.

All buttons or only the exist one?

knirirr:
Just the exit button, it seems.

knirirr (approved):

PR #686, opened 2021-02-04
Additional fields 575

knirirr:
This PR is for editing various additional fields that would be found in fairsharing_record.metadata. I've only implemented one, access_points, to demonstrate how the system would work and will add others with separate PRs once this one (already on the large side) has been merged. 
The list of pending additions can be found at :
https://github.com/FAIRsharing/FAIRsharing-API/issues/231

A particular concern is that various sections of the edit form may be working on metadata simultaneously, so some means is needed of making sure that additional_information doesn't clobber changes elsewhere. This is perhaps something for discussion as I'm not sure my solution is the best. Meanwhile, please look in particular at updateAdditionalInformation() in record.js. This makes an additional query to the server to get only metadata, updates it with the changes to access_points, and then performs an update. 

Various sections (such as getting allowedFields on mount in EditAddtionalInfo.vue) will be updated when new sections are added.


src/components/Editor/AdditionalInformation/AccessPoints.vue
```diff
@@ -0,0 +1,247 @@
...
+      return this.getSection('additionalInformation');
+    },
+    initialFields(){
+      return this.getSection("additionalInformation").initialData
+    },
+    currentFields(){
+      let data = this.getSection("additionalInformation").data
+      if (!data.access_points) {
+        data.access_points = [];
+      }
```
terazus:
```suggestion
      if (!data.access_points) data.access_points = [];
```
In JS you are allowed to write single statement of an IF/Else on the same line without the curly brackets.

src/components/Editor/AdditionalInformation/EditAdditionalInfo.vue
```diff
@@ -0,0 +1,166 @@
...
+    },
+    message(){
+      let error = this.getSection("additionalInformation").error;
+      return {
+        error: error,
+        value: this.getSection("additionalInformation").message,
+        type: function(){if (error){return "error"} else {return "success"}}
+      };
+    }
+  },
```
terazus: I don't think you need this, the Alert component is handling that for us.

src/components/Editor/AdditionalInformation/AccessPoints.vue
```diff
@@ -0,0 +1,247 @@
...
+  name: "AccessPoints",
+  data() {
+    return {
+      openEditor: false,
+      newAccessPoint: {},
+      rules: {
+        isRequired: function(){return isRequired()},
+        isUrl: function(){return isUrl()},
+      },
+      availableTypes: ["REST", "SOAP", "SPARQL", "Other"]
```
terazus: We may want to put this in the editor store for:
1- reusability
2- consistency
3- extendability

src/components/Editor/AdditionalInformation/AccessPoints.vue
```diff
@@ -0,0 +1,247 @@
...
+      }
+      return data;
+    },
+    message(){
+      let error = this.getSection("additionalInformation").error;
+      return {
+        error: error,
+        value: this.getSection("additionalInformation").message,
+        type: function(){if (error){return "error"} else {return "success"}}
+      };
```
terazus: I don't think you need this, the Alert component is handling that for us (in the parent component)

src/components/GraphClient/queries/getMetadata.json
```diff
@@ -0,0 +1,8 @@
+{
```
terazus: Maybe rename this file getRecordMetadata.json and put it into queries/editor/

src/store/record.js
```diff
@@ -32,8 +33,9 @@ let recordStore = {
         },
         sections: {
             generalInformation: initEditorSections(false, ["generalInformation"]).generalInformation,
+            organisations: {},
+            additionalInformation: initEditorSections(false, ["additionalInformation"]).additionalInformation,
             publications: initEditorSections(false, ["publications"]).publications,
```
terazus: You shouldn't need to do this. I was doing it for the generalInformation tab because it's also re-used by the "create new record" page and needed to enforce 2 way dsata binding by pre-declaring the fields.
Look how both data_access and organisations are empty on dev.

src/store/record.js
```diff
@@ -298,6 +300,39 @@ let recordStore = {
...
         },
+        async updateAdditionalInformation({ state, commit}, options){
+            commit("resetMessage", "additionalInformation");
+            let access_points = state.sections.additionalInformation.data.access_points;
+            /*
+                What this does:
+                1. Get metadata (new GraphQL query)
+                2. Modify with access_points,
+                3. Post again (restClient.updateRecord).
+             */
```
terazus: You don't need to do this. You can make a deep copy of state.generalInformation.data.metadata.
This will spare an extra not needed query. Make sure it's an hard copy though.

src/store/record.js
```diff
@@ -298,6 +300,39 @@ let recordStore = {
...
+            let response = await restClient.updateRecord(record);
+            if (response.error) {
+                commit("setSectionError", {
+                    section: "additionalInformation",
+                    value: response.error
+                });
+                return response.error;
+            }
+            else {
+                commit("setMessage", {target: "additionalInformation", value: "Record successfully updated!"});
```
terazus, line 343: We need a point of discussion here: updating the other sections that rely on metadata to avoid overwritting when saving.

terazus (changes requested):

src/components/GraphClient/queries/getMetadata.json
knirirr: This could be deleted, if it's not needed as discussed below.

src/store/record.js
```diff
@@ -300,6 +319,34 @@ let recordStore = {
...
+            newRecord.metadata.access_points = state.sections.additionalInformation.data.access_points;
+
+            /// below = dodgy
+            // define record
+            const record = {
+                record: newRecord,
+                token: options.token,
+                id: options.id
+            };
+            let response = await restClient.updateRecord(record);
```
terazus:
```suggestion
            let response = await restClient.updateRecord({
                record: newRecord,
                token: options.token,
                id: options.id
            });
```

src/store/record.js
```diff
@@ -122,6 +126,22 @@ let recordStore = {
             state.sections.dataAccess.changes = 0;
             state.sections.dataAccess.message = "Record successfully updated!";
         },
+        setAdditionalInformation(state, additionalInformation) {
+            let record = {
+                access_points: additionalInformation['access_points'],
+            };
+            // TODO: Separate setting necessary for each available field...
+            ['access_points'].forEach((type) => {
```
terazus:
```suggestion
           Object.keys(additionalInformation).forEach((type) => { ... }
```
OR
```suggestion
           Object.keys(record).forEach((type) => { ... }
```

src/store/record.js
```diff
@@ -122,6 +126,22 @@ let recordStore = {
...
+                state.sections.generalInformation.data.metadata[type] = JSON.parse(JSON.stringify(record[type]));
+                state.sections.generalInformation.initialData.metadata[type] = JSON.parse(JSON.stringify(record[type]));
+                state.sections.dataAccess.data.metadata[type] = JSON.parse(JSON.stringify(record[type]));
+                state.sections.dataAccess.initialData.metadata[type] = JSON.parse(JSON.stringify(record[type]));
```
terazus: I don't think this is needed. DataAccess will take the extra metadata it needs from generalInformation.initialData. So you are setting something that isn't used.
See `updateDataAccess()`: https://github.com/FAIRsharing/fairsharing.github.io/blob/1ed1bf3052348b321c801c9ed98971b39f9d2e9e/src/store/record.js#L352-L354

src/store/record.js
```diff
@@ -114,6 +116,8 @@ let recordStore = {
             };
             state.sections.generalInformation.data.metadata.support_links = JSON.parse(JSON.stringify(record.support_links));
             state.sections.generalInformation.initialData.metadata.support_links = JSON.parse(JSON.stringify(record.support_links));
+            state.sections.additionalInformation.data.metadata.support_links = JSON.parse(JSON.stringify(record.support_links));
+            state.sections.additionalInformation.initialData.metadata.support_links = JSON.parse(JSON.stringify(record.support_links));
```
terazus, line 120: Instead, I would have taken the support links from sections.dataAccess.initial_data. Both work though.

terazus (changes requested):

src/store/record.js
knirirr: I'm not quite sure what you mean - would you mind giving an example, please?

src/store/utils.js
```diff
@@ -78,6 +78,7 @@ export function initEditorSections(data, sectionsNames){
                 }),
                 is_dataset: false
             },
+            additionalInformation: JSON.parse(JSON.stringify(data.metadata)),
```
terazus, line 81: Not the most elegant but fine :). I would only have taken the fields I need using destructuring instead of copying the full metadata object. A bit more optimized but not a major thing.

src/store/record.js
terazus: I would have removed these two line and added one line here https://github.com/FAIRsharing/fairsharing.github.io/blob/b2ad7cf5c86f7eef50bf4a2b639e21a2ea8fe740/src/store/record.js#L325
` newRecord.metadata.support_links = state.sections.dataAccess.initialData.support_links ;`
Both work though.

terazus (approved):

knirirr:
I'll leave the branch open for further work on additional fields on this area of the forms.

PR #692, opened 2021-02-08
Bizarre javascript date behaviour.

knirirr:
Tests doing something odd here, but I think this should sort it out.


ramGranell (changes requested):
To be coherent I suggest to change also other lines in this file that appear the same code:
      auxString = date.toLocaleString('default', { month: 'short' })+' '+date.getUTCDate()+ ', '+date.getUTCFullYear();

terazus (approved):

ramGranell (approved):

PR #697, opened 2021-02-09
Update Organisation with new layout and functionality

hoseinmirian:
new organizations layout:

- Smooth link hover effect
- showing grants and type of organization

for testing a record with multiple organizations please use 👍 

for testing with no organizations use:
http://localhost:8080/#/45


src/components/Records/Record/Organisations.vue
```diff
@@ -2,126 +2,148 @@
...
+/* TODO: Replace with query from database */
+import organisationRelations from '@/data/organisationRelations.json';
+export default {
+  name: "Organisations",
+  components: {
+    SectionTitle
+  },
+  data(){
+    return {
+      /* TODO: Replace with query from database */
```
knirirr: What's happening with this TODO, and the one at line 78?

knirirr (changes requested):
In general I think the layout looks good. However, there's one special case that's ought to be handled differently, which is multiple grants from the same provider, e.g. 

https://deploy-preview-697--fairsharing.netlify.app/#/3144

It would be better to consolidate all grants into the same list.

Screenshot 2021-02-11 at 14 12 59

src/components/Records/Record/Organisations.vue
hoseinmirian: I don't know these were left from past. I did not touch those parts

knirirr: In which case they are probably left behind from old work, and would need to be removed.

hoseinmirian:
image
duplicate organisation is deleted and if there are 2 grants or more for one organisation all are combined into one array and shown as grant(s) for multiple items and grant for single item

src/components/Records/Record/Organisations.vue
```diff
@@ -2,126 +2,169 @@
...
-          </v-tooltip>
-          <p class="ma-0">
-            {{ organisationLink.grant.name }}
-          </p>
-        </div>
-      </v-card>
+                  <strong v-if="grant">
+                    {{ grant.name }}
+                  </strong>
+                  <span v-if="grantIndex!==organisationLink.grants.length-1"> , </span>
```
knirirr: It's a very minor thing, but would you mind removing the space before the comma, please, i.e.:

```
<span v-if="grantIndex!==organisationLink.grants.length-1">, </span>
```

knirirr (approved):

src/components/Records/Record/Organisations.vue
hoseinmirian: okay it will be done in next commits

Issue #655, opened 2021-01-18
Redesign Record display


hoseinmirian:
- [x] General Information
- [x] Publication
- [x] Related Content
- [x] Collections
- [x] Support
- [x] DataConditions
- [x] Tools
- [x] Organisations
- [x] Record Restructuring
- [x] Spliting up the prs

PR #702, opened 2021-02-10
Hossein data condition #655

hoseinmirian:


hoseinmirian:
Please ignore the way icons look due to they are easy to be changed from the icons.js

Different case scenario to test:

http://localhost:8080/#/57 
- for seeing data_versioning

http://localhost:8080/#/FAIRsharing.6xq0ee
- showing all possible sections 4sub sections (data curation, data access, data release, data versioning)

http://localhost:8080/#/16
- showing only data access

http://localhost:8080/#/12
- deactivating section when no data available

http://localhost:8080/#/4
- the section is active even with one license

http://localhost:8080/#/FAIRsharing.3kcgmr
- no licenses and no data_processes available

hoseinmirian:
Also in my last commit, I have covered a very rare case scenario but possible where the type of the data_processes's object is not defined in icon.js like below and still page will load that type using an undefined icon:
for e.g object comes from API like {name:'',URL:'',type:'not defined! or any name'} then the front end will show it the way below:
So in order to not face this issue next time please take the below steps:
1- make sure for each type you define in the back-end or curator's dashboard, there is an equivalent key defined with underscore with the name of type in icon.js like : {type:'data access'} in icon.js it will be defined as data_access. (using an underscore)
2- if the new data type is defined in back end but its icon is not defined in icon.js still we can see an undefined icon and data shown nicely in the page. based on that behavior of layout we know that we have not defined the equivalent of new data type of back-end in front-end and immediately will take a look at icon.js and add the new underscored type name. 

image

knirirr (changes requested):
The provided links demonstrate things nicely.

Concerning licences, they're being displayed with simply the licence name. However, their relation to the record is also available and it's likely this will be requested. It's available via a different field. The licences field is thus:
```
"licences": [

    {
        "id": 143,
        "name": "Creative Commons Attribution 4.0 International (CC BY 4.0)"
    }

],
```

However, there's also a licenceLinks field that shows the relation between the record and each licence:

```
"licenceLinks": [

    {
        "relation": "undefined",
        "id": 2448,
        "fairsharingRecord": {
            "id": 1453
        },
        "licence": {
            "id": 143,
            "name": "Creative Commons Attribution 4.0 International (CC BY 4.0)"
        }
    }

],
```

Using this field instead would provide the necessary relation information.

hoseinmirian:
record No.158(http://localhost:8080/#/158) is shown as below. currently, most of the record relation value is undefined. when any other data is provided except undefined, it will show relation:"type of relation" below each licence.

image

knirirr (changes requested):
The code seems to be working as expected.
I would request one small change, though. 
Looking at the output, licence relationships are shown as snake case strings, e.g.

Screenshot 2021-02-12 at 13 32 40

I'm happy with that, but I expect users might find it a little odd. Therefore, it might be worthwhile replacing the underscores with spaces on the page only.
There's already a `cleanString` function in utils/stringUtils.js so this should be simple enough.

hoseinmirian:
image

its done. final commit

knirirr (approved):

PR #711, opened 2021-02-15
Hossein new support #655

hoseinmirian:
different test case:

- http://localhost:8080/#/FAIRsharing.9qv71f
contains some new sub sections. nice to see the icon allocated for them

- http://localhost:8080/#/FAIRsharing.6k0kwd
same explanation as above

http://localhost:8080/#/1 
having only contacts available with 3 attributes such as user_email, user_name, and user_orcid

http://localhost:8080/#/2000
a record with new subsections in support where their icons are not defined but it does not break the layout and shows undefined icon which is question mark. 

http://localhost:8080/#/300
a record with no support active


knirirr (changes requested):
There's some odd stuff going on with the icons here; for example, a question mark rather than an envelope for email. Also, the TeSS links show a Facebook and training documentation for Twitter. The first of these ideally needs to be an Elixir logo and I'm not sure about the latter, but definitely not Twitter.

Screenshot 2021-02-15 at 15 26 51

Re. the TeSS logo, here's what it ought to be:

https://www.jalview.org/file/328

But, there could be problems getting that to match our existing theme.

hoseinmirian:
@knirirr thanks for your comments. for icons its easy so I can update the appropriate icons. In regard to elexir icon tess training, I am wondering if I am allowed to turn its icon into a black and white icon so it fits with our design? otherwise we have a harder option like asking font-awesome to provide us this icon. Another option is to not provide icons at all and show as undefined.

knirirr:
> In regard to elexir icon tess training, I am wondering if I am allowed to turn its icon into a black and white icon so it fits with our design?

That would seem like the best option, if you can do it.
Technically it's Elixir UK but you may not be able to fit the UK part in.

hoseinmirian:
please check the below link arrayExpress:
- icons are now relevant to the topic
- now we have a CustomIcon that can be used within icon.js by defining a source key.. (now we can import any icon from external resources) 
- a new custom_icons folder is created in assets for future icons we need to add there. 


image

knirirr (approved):

PR #718, opened 2021-02-16
edit watched records only for curators

terazus:
Quick fix to only show the edit button of the watched records for curators.


knirirr:
Unfortunately I'm seeing an issue on this branch; logging out when on the profile page throws this:

```
TypeError: can't access property "maintenanceRequests", this.user().records is undefined
    records RecordCard.vue:77
...
```

There's also:

```
[Vue warn]: Error in render: "TypeError: can't convert undefined to object"
```

If a different panel is selected.

knirirr (approved):

PR #720, opened 2021-02-17
Dom fix edit bugs

terazus:
Fixed two bugs:

- Couldn't save publications due to missing record id passed in the restclient.
**Solution**: add the id to the object passed to the RESTClient method.

- Couldn't save data access because it was trying to associate values to a state that wasn't set yet.
**Solution**: Comment lines 119 and 120 that weren't actually doing anything. AdditionalInformation builds the metadata object from general information (Line 324), which is updated when data access is updated. (Lines 117 and 118)


src/store/record.js
```diff
@@ -116,8 +116,8 @@ let recordStore = {
             };
             state.sections.generalInformation.data.metadata.support_links = JSON.parse(JSON.stringify(record.support_links));
             state.sections.generalInformation.initialData.metadata.support_links = JSON.parse(JSON.stringify(record.support_links));
-            state.sections.additionalInformation.data.metadata.support_links = JSON.parse(JSON.stringify(record.support_links));
-            state.sections.additionalInformation.initialData.metadata.support_links = JSON.parse(JSON.stringify(record.support_links));
+            //state.sections.additionalInformation.data.metadata.support_links = JSON.parse(JSON.stringify(record.support_links));
+            //state.sections.additionalInformation.initialData.metadata.support_links = JSON.parse(JSON.stringify(record.support_links));
```
knirirr: Could these simply be deleted, if they're not needed?

terazus:
```suggestion
```

knirirr (approved):

PR #722, opened 2021-02-17
Fix getter

### LINKED ISSUE #700: getField refactoring in record.js
This function needs refactoring: 

https://github.com/FAIRsharing/fairsharing.github.io/blob/additional_fields_575/src/store/record.js#L405-L412

1. To prevent going through the loop for each field, i.e. only if tags includes fieldName.
2. Avoid setting values in the getter, e.g. a temporary variable for the result could be created and returned.

hoseinmirian:


hoseinmirian:
the task focus on not looping over each fieldName and instead check if the fieldName is within the tags array so it improves the performance of the getter..

knirirr (approved):

PR #724, opened 2021-02-17
Dom improve user profiles

terazus:
Improves the user private profiles page.
Also added a loader to search pages and move the autcomplete filter id autoscroll to avoid fields being partially hidden.


knirirr (changes requested):
Looks good, though it would be useful to add back in the option to edit a watched record if the logged-in user is a curator.

knirirr (approved):

PR #725, opened 2021-02-18
Additional fields 575

knirirr:
This adds associated_tools to the additional information panel; it's pretty much the same as access_points.


src/components/Editor/AdditionalInformation/AssociatedTools.vue
```diff
@@ -0,0 +1,218 @@
...
+  >
+    <v-card-title>
+      Associated Tools
+    </v-card-title>
+    <v-card-text>
+      <p
+        v-if="currentFields.associated_tools.length === 0"
+        class="my-0 mx-3"
+      >
+        No access points defined.
```
terazus:
```suggestion
        No Associated tools defined.
```

terazus (approved):
You are missing subform validation for both elements (the ones in the overlay). This means you can add a completly empty object:
validate

terazus (changes requested):
Error from approving previous request

terazus (approved):

Issue #726, opened 2021-02-18
Additional info. component validation


knirirr:
As mentioned here:

https://github.com/FAIRsharing/fairsharing.github.io/pull/725#pullrequestreview-593207308

The subcomponents of the additional info. field need validation.

PR #727, opened 2021-02-18
Data processes 575

knirirr:
This adds data_processes and also refactors an array comparison function into utils. I'll apply this to the other components in a separate PR.

Similarly, this needs fixing: https://github.com/FAIRsharing/fairsharing.github.io/issues/726, probably as part of refactoring.


terazus (changes requested):
Also: associated tools changes updated isn't correctly emmited but it may be bound to your other bug.

PR #732, opened 2021-02-23
Dom cleaning stores

terazus:
Cleaned up users, record and editor states when leaving pages where they are required.


knirirr (approved):

src/store/record.js
```diff
@@ -163,6 +160,7 @@ let recordStore = {
             }
         },
         cleanRecordStore(state){
+            state.sections = null;
```
hoseinmirian, line 163: here webstorm warns. I changed the value to undefined or {} and the error was gone. 
image

hoseinmirian (approved):

PR #734, opened 2021-02-23
Fix bug with undefined licence id

### LINKED ISSUE #731: Licence relation bug in DataAccess.vue
Some records (e.g. 10.25504/FAIRsharing.9k7at4) produce the following console errors when loading, which appears to be related to looking for a licence relation when licences are not defined.

Screenshot 2021-02-23 at 11 35 16

hoseinmirian:


hoseinmirian:
some record have licence with undefined licence ID and that was the reason we were getting errors. now its fixed 
relevant record:
http://localhost:8080/#/FAIRsharing.9k7at4

knirirr (approved):

PR #746, opened 2021-03-02
Hossein clean store #729

### LINKED ISSUE #729: Store memory leakage
The store needs to be cleared upon component destruction, e.g. after leaving the search and edit pages. 

- [x] records.js @hoseinmirian 
- [x] editor.js  @terazus 
- [x] record.js @terazus 
- [x] users.js  @terazus

hoseinmirian:
Clean records store before destroy  ...


terazus (changes requested):
Even though the coverage is 100% you may be missing an expect statement in your records.spec.js test suite.
You want to make sure that when the component is destroyed the states have been correctly reset so something like:
`expect(recordsStore.records.length).toBe(0);
expect(recordsStore.facets.length).toBe(0);
["hits, "loading", ...].forEach(val => {
    expect(recordsStore[val]).toBe(null);
})`
Alternativerly, instead of expecting from the store, you can expect a computed property, if bound through a mapState:
`expect(wrapper.vm.records.length).toBe(0)`

terazus:
> Alternativerly, instead of expecting from the store, you can expect a computed property, if bound through a mapState:
> `expect(wrapper.vm.records.length).toBe(0)`

Actually no, you can't do that sorry. Because the component should be destroyed, wrapper.vm shouldn't be available anymore.

knirirr (approved):

terazus (approved):

PR #767, opened 2021-03-04
Dom edit feedback

terazus:
Contains the modifications of the general information tab after the second round of review.


terazus:
Wrong target branch

PR #778, opened 2021-03-08
Public user profile

knirirr:
User profiles will be available at `/users/:id` and should be visible even when logged out.
If you have set `hide_email` to true then your email address won't be visible.

N.B. I have used the User.vue file as a basis and removed un-necessary items as I spotted that there's already a component for tables and so didn't think creating several more components was a good idea. To do this, the "userMeta" field that previously came from the logged-in user's data must now be obtained from the same query to the API that gets maintained records.


ramGranell (approved):

terazus (approved):

terazus (changes requested):
I think we need to handle 404 when user is not found.
Example:
image

terazus (approved):

PR #779, opened 2021-03-09
Dom centralize icons

terazus:
This is a wrapper from centalized icons inside vuetify. Just pass your icon source name (taken from @/plugins/icons.js) as the item property of the wrapper and it should work. If it doesn't find the item it will rely on a default fallback icon that you can customize.
If this acceptable to everyone I will extend the edit system and curation dashboard to use the new Icon component.

It handle both images and icons.

Example of usage:
To get a small blue github icon without the dashed circle container:
`
<Icon
      item="github"
      wrapper-class=""
      size="small"
      color="primary"
/>
`

To get the "collection" image inside a dashed circle and customize the size
`
<Icon
      item="collection"
      height="80"
/>
`

To get a small database image without the dashed circle with a custom falling back to the 'fa-handshake' icon mapped from the collaborates_on key (default fallback is mapped to the 'undefined' string key and points to a question mark icon).
`
<Icon
      item="database"
      wrapper-class=""
      height="30"
      fallback="collaborates_on"
/>
`

You can also directly access the vuetify icon inside an HTML template using `this.$vuetify.icons[iconName]` (default) or `this.$vuetify.icons[iconName].icon` (customs)


src/components/Records/Shared/RecordStatus.vue
```diff
@@ -9,18 +9,18 @@
...
           v-on="on"
         >
-          <img
-            :src="('./' + recordType[record.type].icon)"
-            :alt="getRecordStatus.title"
-          >
+          <Icon
+            :item="record.type"
+            wrapper-class=""
+            :height="80"
```
hoseinmirian, line 19: here if change the value of height to either lower like 60 or even higher like 100 the size of the icon remains the same.. 
based on the design the inner circle must be a little bit smaller than the dashed circle border. 

dev version
image


this branch version
image

hoseinmirian (changes requested):

src/components/Records/Shared/RecordStatus.vue
terazus: Im trying to fix it but there's so much css copied pasted all over the place that it's very very very hard to figure it out.

terazus:
Be careful not to exceed the container size in RecordStatus or the image will not overflow and will get cropped.
image

knirirr (approved):

ramGranell (approved):

Issue #752, opened 2021-03-03
Organisation Page


knirirr:
As mentioned at https://docs.google.com/document/d/1g1sp9vSFEfkLw6pyiESmjdxRRUiN9tjiz8k7Gbv7_VE/edit?ts=603f88be

"...where will the link to the organisation page on FAIRsharing be? (so users can see all the records associated with an organisation)"

Further information (from Zulip):

> In terms of fields, we would need:
> 
>     - Logo
>     - Name
>     - Alternative names
>     - Type
>     - Child organisations
>     - Parent organisations (so users can click up, say from OERC to Dept. of Engineering Sci)
>     - Related records and their associated grants/status
>     - (non-MVP) ROR cross ref ID

This last field cannot be done at present (not supported in the API) and will require a separate ticket at a later date.

PR #783, opened 2021-03-10
Organisations page

knirirr:
Currently at /organisations/:id, fields as per #752 - I had to add some more fields to the responses from the server.
I've not done any UI elements, just queried the server for the data.


knirirr:
Example: /#/organisations/1

knirirr:
The example organisation (no. 1) as mentioned above could be used for viewing (https://deploy-preview-783--fairsharing.netlify.app/#/organisations/1 at the time of writing).
I've changed the links to parent and child organisation as it seems that Vue router doesn't like to reload the same component, and simply pushing a new route will (as done on the record page) will fail.

ramGranell (approved):

hoseinmirian (approved):
It looks good. One suggestion I would say is to add a watcher to make the page reactive when ID of the organisation changes. Could happen when either the browser address is updated manually or by pressing the back button. 

image

src/views/Organisations/Organisation.vue
```diff
@@ -68,133 +58,146 @@
...
-                <v-list-item
-                  :key="'organisation_types'"
-                  class="body-1"
-                >
-                  <v-list-item-content
-                    class="py-0 d-block"
+                    </v-list-item-content>
+                  </v-list-item>
+
+                  <!-- Parent organisations -->
```
hoseinmirian: can be changed into child organisation. its duplicate

hoseinmirian (commented):
> 
> 
> The example organisation (no. 1) as mentioned above could be used for viewing (https://deploy-preview-783--fairsharing.netlify.app/#/organisations/1 at the time of writing).
> I've changed the links to parent and child organisation as it seems that Vue router doesn't like to reload the same component, and simply pushing a new route will (as done on the record page) will fail.

please ignore this comment.. 

image

hoseinmirian (changes requested):
please apply the changes. tnx

src/views/Organisations/Organisation.vue
hoseinmirian: in Maintainer.vue an eslint issue spotted. 
image

knirirr:
> It looks good. One suggestion I would say is to add a watcher to make the page reactive when ID of the organisation changes. Could happen when either the browser address is updated manually or by pressing the back button.


Please would you provide an example of what you mean - thanks.

hoseinmirian:
@knirirr 

like currentRoute in Record.vue. it should refetch the organisation data when route changes. 

image

hoseinmirian (approved):

src/router/routes.js
```diff
@@ -15,6 +15,7 @@ export const RequestNewPassword = () =>  import(/* webpackChunkName: "newPwd-chu
 export const ResetPassword = () =>  import(/* webpackChunkName: "resetPwd-chunk" */ "@/views/Users/ResetPassword");
 export const EditProfile = () =>  import(/* webpackChunkName: "editProfile-chunk" */ "@/views/Users/EditProfile");
 export const OauthLogin = () =>  import(/* webpackChunkName: "Oauth-chunk" */ "@/views/Users/Login/OauthLogin.vue");
+export const Organisation = () =>  import(/* webpackChunkName: "Oauth-chunk" */ "@/views/Organisations/Organisation");
```
terazus: You need to be carefull with the webpackChunkName comment which control the chunk name.
```suggestion
export const Organisation = () =>  import(/* webpackChunkName: "Organisation-chunk" */ "@/views/Organisations/Organisation");
```

src/utils/stringUtils.js
```diff
@@ -1,3 +1,4 @@
+const lf = new Intl.ListFormat('en');
```
terazus: This is only used once so it can be declared in the formatList filters function (between L37 and L38)

src/views/Organisations/Organisation.vue
```diff
@@ -0,0 +1,367 @@
...
+        {text: 'Relation', value: 'relation', align: 'center'},
+        {text: 'Grant', value: 'grant', align: 'center'},
+        {text: 'Actions', value: 'actions', align: 'center', sortable: false}
+      ]
+    }
+  },
+  computed: {
+    ...mapState('users', ['user']),
+    currentRoute() {
+      return this.target || this.$route.params['id'];
```
terazus: not sure what this.target refers to, it's never used nor declared. Something you copy pasted from another component ?
```suggestion
      return this.$route.params['id'];
```

terazus (changes requested):

src/views/Organisations/Organisation.vue
knirirr: Isn't it related to the record preview overlay? It's probably come in by mistake from that.

Issue #781, opened 2021-03-10
Curation panel, change icon uploading


ramGranell:
The icons used in the curator panel need to be uploaded using new component https://github.com/FAIRsharing/fairsharing.github.io/pull/779 when branch is merged into dev

Issue #773, opened 2021-03-05
Public user profile


knirirr:
This could be fairly similar to the private one but with the following sections removed:

- Preferences.
- Email address (unless hide_email: false).
- Watched records.
- Created records.
- Maintenance requests.

PR #786, opened 2021-03-11
Rg curation panel#773 #781

ramGranell:
This ticket is for sorting out two things:
#773: urls for user public pages where users exist
#781 : using icons with new loading system
The visual results are in the curator panel (https://deploy-preview-786--fairsharing.netlify.app/#/accounts/login?goTo=%2Fcurator) where there are links to all the user names and the records icons are used


src/components/Curators/MaintenanceRequests.vue
```diff
@@ -39,25 +39,25 @@
...
-                      :src="'./' + recordType[props.item.type].icon"
-                    >
-                  </v-avatar>
-                </span>
                 {{ props.item.recordName }}
               </a>
             </td>
             <td>
-              {{ props.item.userNameID }}
+              <a :href="'#/users/' + props.item.userId">
```
knirirr, line 58: I'm not sure if the # is required here. Do you know, @terazus ?

terazus: It is required as long as we are not in history mode.

terazus (approved):

src/components/Curators/MaintenanceRequests.vue
knirirr: Are we to remain in history mode for deployment? If not then this will have to be reverted later.
If so then some will have to be added (see https://github.com/FAIRsharing/fairsharing.github.io/issues/804).

terazus: We will have history mode in production, which means that we will need to removed all href '#'.
The easiest way i found to deal with this is to use a `goTo(link) { this.$router.push({path:link}) }` method instead bound through a @click. Because $router already knows which mode we are on.

knirirr: If so then it might well be best to sort these links out now so we don't have to remember to come back to them later (hence #804).

ramGranell: Well, I've been creating all the links in the curator panel in this way (e.g. https://github.com/FAIRsharing/fairsharing.github.io/blob/dev/src/views/Curators/Curator.vue#L68). Maybe I can create a ticket to modify them when real development happens.

knirirr: @ramGranell please create a ticket, then; I could approve this once I know the link issue has been noted.

ramGranell: I created the ticket https://github.com/FAIRsharing/fairsharing.github.io/issues/805 but not sure if I explain it properly.

knirirr (approved):

PR #789, opened 2021-03-11
Dom merging relations

terazus:


knirirr (changes requested):
Looks good, but I spotted a couple of things:
1. As should be the case, only policies can recommend, but only collections should be able to collect and it seems that anything can. Do you need me to modify anything from the API to enable that to be enforced?
2. It would be worthwhile excluding self from the list of records available to link - I was able to attempt to link a record to itself (which the API prohibits):

Screenshot 2021-03-15 at 11 31 04

knirirr (approved):

knirirr:
A small change to this branch and two modifications to the API have addressed the issues mentioned above:

https://github.com/FAIRsharing/FAIRsharing-API/issues/259
https://github.com/FAIRsharing/FAIRsharing-API/issues/258

PR #794, opened 2021-03-15
refactored support links into a grouped table

terazus:


knirirr:
Screenshot 2021-03-17 at 10 51 44
I noticed something a bit odd going on whilst looking at this, specifically I can't add help documentation. It appears that:

1. The name isn't being sent to the API.
2. The API isn't being validated correctly.

...this also causes the buttons to remain stuck on loading. 
I've fixed 2 with a schema change and will be pushing this shortly. I can confirm that 1 is still a problem even with the fix.

knirirr (changes requested):
I have now pushed the schema changes which will fix this issue.

terazus:
Names should now be properly submitted.

knirirr (approved):

PR #801, opened 2021-03-16
Allow hidden records to be viewed by admins.

knirirr:
Previously, the call to fetchRecord in Record.vue didn't pass the user's JWT, and so the relevant API controller always had current_user set to nil. The result was that even a curator couldn't see a hidden record (they should see it, but with a warning banner). 
Also, this required hiding the action menu if there's an error on the record page, otherwise it's displaced by the 404 message.


hoseinmirian (approved):

src/store/record.js
```diff
@@ -191,11 +191,14 @@ let recordStore = {
...
     actions: {
-        async fetchRecord(state, id){
+        async fetchRecord(state, options){
             state.commit("resetCurrentRecordHistory");
             recordQuery.queryParam = {
-                id: id
+                id: options.id
             };
+            if (options.token) {
+                client.setHeader(options.token);
```
terazus, line 200: You will need to reset the headers once you are finished with the query.
You can use `client.initalizeHeader()` after the awaits.
Otherwise the jwt will stay in memory as an attribute of the header untill it's being reset.

terazus (changes requested):

PR #820, opened 2021-03-22
Dom graphs

terazus:


knirirr:
I've changed my graph PR (#830) to go to dev. so you could now close this if you wished.

PR #821, opened 2021-03-22
Dom massi edit additional information

terazus:


terazus:
Dunno why the coverage report in the checks won't update but its here: https://coveralls.io/builds/38642374

knirirr:
I noticed that the "save and continue" and "save and exit" buttons both act like the former on this tab.

tests/fixtures/additionalInformation.json
```diff
@@ -0,0 +1,303 @@
...
+            "default": "data access",
+            "description": "Data access is used when the process is accessing data, e.g. searching, browsing, viewing, and/or visualizing the data. Data curation is used when the process provides data curation tasks, e.g. submitting and/or editing the data. Data release is used when the process is releasing data, e.g. downloading and/or exporting the data. Data versioning is used when the process is versioning data, e.g. upload of new versions of data."
+          }
+        }
+      }
+    },
+    "description": "Properties specific to repositories",
+    "additionalProperties": false
+  },
+  "data": {
```
knirirr, line 297: Is this something left over from debugging, or is it used somewhere?

knirirr (changes requested):
This looks good to me, though there's one part of the field definitions I didn't understand (see comment) and the save button will need dealing with.

tests/fixtures/additionalInformation.json
terazus: This is a test fixture im using to run the schemas tests. What's the problem?

knirirr: None. I couldn't spot where it was used and wondered about it.

terazus:
Redirection should now behave correctly

knirirr (approved):

PR #822, opened 2021-03-22
Rg stats first page

ramGranell:
The statistics can be seen here: https://deploy-preview-822--fairsharing.netlify.app/#/summary-statistics
There is an example of a pie chart and a bar chart and their current layout has been approved. The 29 missing charts will be created in the same way that these two charsts.


knirirr:
Is it supposed to be blank when anything other than "all" is selected?

ramGranell:
> Is it supposed to be blank when anything other than "all" is selected?

Well, this ticket it was to show the layout of two charts and about how to process the data for the API. The button selection functionality will be implemented when the 31 charts were implemented. At the moment there are just two charts that are not standard/databases/policies specific and for this reason they do not appear when these options are selected.

knirirr (approved):

PR #826, opened 2021-03-24
Hossein json ld #818

### LINKED ISSUE #818: Add markup to record pages
I've added add additional field to fairsharingRecordType, as per this ticket: https://github.com/FAIRsharing/FAIRsharing-API/issues/260

This field ('schemaOrg') can be added to the query here:

https://github.com/FAIRsharing/fairsharing.github.io/blob/master/src/components/GraphClient/queries/getRecord.json

Then, the output need only be put into the header of the page. Any page on the old site should provide an example, e.g.

http://fairsharing.org/bsg-d000001

```
<script type="application/ld+json">
{
        "@context": "http://schema.org", 
        "@id": "https://doi.org/10.25504/FAIRsharing.9kahy4", 
        "@type": "Dataset", 
        "alternateName": "GenBank", 
        "citation": {
                "@type": "CreativeWork", 
                "identifier": "doi:10.25504/FAIRsharing.9kahy4"
        }, 
        "description": "This FAIRsharing record describes: GenBank is the NIH genetic sequence database, an annotated collection of all publicly available DNA sequences. The complete release notes for the current version of GenBank are available on the NCBI ftp site. A new release is made every two months. GenBank growth statistics for both the traditional GenBank divisions and the WGS division are available from each release. GenBank is part of the International Nucleotide Sequence Database Collaboration (INSDC), which comprises the DNA DataBank of Japan (DDBJ), the European Molecular Biology Laboratory (EMBL), and GenBank at the NCBI. These three organizations exchange data on a daily basis. ", 
        "identifier": "10.25504/FAIRsharing.9kahy4", 
        "license": "https://creativecommons.org/licenses/by-sa/4.0/. Please link to https://fairsharing.org and https://fairsharing.org/static/img/home/svg/FAIRsharing-logo.svg for attribution.", 
        "name": "FAIRsharing record for: GenBank", 
        "url": "https://doi.org/10.25504/FAIRsharing.9kahy4"
}
</script>
```

hoseinmirian:
Json LD data is added in home page in a static way. 

as screenshot below shows json ld data is added when we check head tag from the inspector tool 

image


knirirr (changes requested):
It looks like there's a mix-up here between two different requirements:

1. There needs to be some JSON-LD markup on the home page, but the content should be almost the same as shown here:

https://github.com/ISA-tools/biosharing/blob/master/biosharing_web/templates/home.html#L31-L102

...although the "identifier" will change as we're not using bsg IDs by default any more.

2. What's been included on the home page here is in fact the data for a single record. Instead, information for each record should be displayed on an individual record's page. Those data will be available in the "schemaOrg" field of the API response (you will probably need to add this field to the relevant query).

terazus:
vue-jsonLD doesn't build on their master branch so I would use another alternative.

hoseinmirian:
@knirirr please ignore the home page test due to its been done on the other PR.

knirirr (approved):
I can find the JSON in the source, though I must do further work on finding a good way for a client to extract it (as users will ask).

Issue #670, opened 2021-01-29
Approval banner


knirirr:
On the record page a banner should be displayed to warn users if a record is hidden or not approved. There's already an is_approved field but an is_hidden field must also be added. Here's an example of what's done in the previous system:

https://github.com/ISA-tools/biosharing/blob/master/biosharing_web/templates/components/viewing/approvals.html

There are two approval options which would need to be combined into one for the new system, showing something like this if is_approved == false:

"This record is awaiting review by FAIRsharing curators."

knirirr:
This banner is still missing on the latest master, after merging dev.

PR #848, opened 2021-04-12
Hossein approval hidden banner #670

hoseinmirian:


src/views/Records/Record.vue
```diff
@@ -19,11 +19,21 @@
...
+            v-if="alreadyClaimed || claimedTriggered || user().is_curator"
             class="d-flex flex-grow-1"
           >
+            <v-alert
+              v-if="user().is_curator && currentRecord.fairsharingRecord['isHidden']"
+              dense
+              type="info"
+              class="mr-3"
+            >
+              <span>the record is hidden</span>
```
knirirr:
```suggestion
              <span>This record is hidden!</span>
```

knirirr (changes requested):
One thing I noticed - if viewing a hidden record when not logged in the 404 message and the action button appear, with the latter displaced. The action button ought to be hidden if the record doesn't show to prevent this. I think I fixed it elsewhere but the problem seems to have crept back in again.
Otherwise, the banner is good.

hoseinmirian:
while being logged in in a hidden record:
image

please refresh each time. 

while being logged out in a hidden record
image

normal behavior of other record for e.g record number 2 when not logged in shows action button
image

when logged in same record: (same output)
image

knirirr (approved):

PR #859, opened 2021-04-13
Remove v-html checking for specific files only, not whole app.

knirirr:
The modified comment blocks should fix this issue, so I've been able to remove the entry in in the eslint config.


terazus (approved):
a much better solution.
So basically we forgot to close the "tag" ?

knirirr:
> So basically we forgot to close the "tag" ?

Yes; I think the tag originally used can only handle ignoring the very next line, but this was a multi-line html tag so the only solution can be to wrap it as done here.

hoseinmirian (approved):
thanks. good way of doing it

PR #863, opened 2021-04-13
Dom filters sm

terazus:
Addind filters on SM and bellow:
resFil


hoseinmirian (changes requested):
1- two scrolls in a screen looks odd when popup is open and cause UX issue. the main scrollbar can be disabled when this popup is on top. 
body{overflow:hidden}

2- left panel Filter panel has lost its stickiness in an above small screen sizes. 

image

knirirr:
I didn't encounter any issues when the filters were open, and they seemed easy enough to use (given the constraints of a small screen). 
However, the left panel no longer scrolls - is that intended, or a side-effect?

terazus:
> 
> 
> I didn't encounter any issues when the filters were open, and they seemed easy enough to use (given the constraints of a small screen).
> However, the left panel no longer scrolls - is that intended, or a side-effect?

I can't reproduce. We can see thr panel scrolling on my gif.

knirirr:
> I can't reproduce. We can see thr panel scrolling on my gif.

Unfortunately, I can't reproduce your gif. If I go directly to https://deploy-preview-863--fairsharing.netlify.app/#/search?fairsharingRegistry=Standard (with a full-sized screen) then the filter panel has no scroll bar and is fixed relative to the search results.

hoseinmirian:
@knirirr @terazus 
there are 2 issues responsible for 2 scroll being open for a mobile user:
- screen size is small and user normally tend to scroll with their fingers and expect one thing to be scrolled at a time
- buttons and filters already taken up space and if user accidentally press any button so we do not want them to also involve with these kind of issues as much as possible. 

@terazus for seeing the left panel is broken only refresh the page in a normal screen and check the left panel stickiness.

hoseinmirian:
> 
> 
> > I can't reproduce. We can see the panel scrolling on my gif.
> 
> Unfortunately, I can't reproduce your gif. If I go directly to https://deploy-preview-863--fairsharing.netlify.app/#/search?fairsharingRegistry=Standard (with a full-sized screen) then the filter panel has no scroll bar and is fixed relative to the search results.

yes because your screen size is large. Please use zoom in ctrl+ plus button to see if its sticky or not.

knirirr:
> yes because your screen size is large. 

Exactly - I am seeing unexpected behaviour on large screens as a result of changing the code to accommodate small ones.

knirirr:
Further testing on an iPad shows that the tools are fine to use - I had no trouble with them. Only one problem showed up which is that tooltips persist after the filters are closed:

IMG_F467DC0FF590-1

knirirr (approved):
I'm happy with this other than the bug mentioned here:

https://github.com/FAIRsharing/fairsharing.github.io/issues/867

I suggest creating a separate PR to fix that.

PR #886, opened 2021-04-15
Dom maintenance mode

terazus:
this one is a bit tricky to test. You need to open introspector store and uncomment line 68 to force maintenance mode.
OR disable your VPN. Any error loading the site will result in a maintenance mode.


knirirr:
Maintenance seems fine, but I'm not noticing anything when I set the Read-Only header on the server. But, attempts to save are correctly met with `{ "message": "Record modification is currently disabled." }` from the server.

terazus:
what could the problem be?
This is what I'm getting from the API:
image

This is what i do in the store:
image

What do you think is wrong ? the uppercase maybe?

knirirr:
It appears to be related to the Puma server's presentation of the headers, so I can see the banner if manually set read_only, but the development webserver wasn't providing the headers (the "public" testing server is another matter).

knirirr (approved):

PR #890, opened 2021-04-15
Hossein licence #813

### LINKED ISSUE #813: Licence static page

hoseinmirian:


knirirr:
Unfortunately this still refers to the old API in the sections "API Access", "API versions" and "API changes". These will all have to be replaced with a single API section making reference to a new API page, which should include this content:

https://github.com/FAIRsharing/FAIRsharing-API/blob/master/README_REST.md

It's Markdown, but that's presumably easy enough to convert to HTML.

knirirr (changes requested):
Removing the API sections as in the previous comments and replacing it with "For information on the API please go to <link>"  (or similar) is needed.

hoseinmirian:
@knirirr thanks for comments. in regard to HTML page if it is static yet its easy but due to I read from a string data in a json file it can be more difficult tho. I need to see if I can make it into a dynamic file reading from json otherwise I just use static HTML code.

knirirr:
A static page is fine here as it's not something that is likely to be changed often.

knirirr:
Now we have a different problem; the rendered markdown looks good, but it must be a separate page from the licence information. I.e. we still need this: https://fairsharing.org/licence
But, without this section, replacing it with a link to the new API instructions:

Screenshot 2021-04-20 at 09 16 04

knirirr (changes requested):
Still not the correct content (see previous comment).

src/views/Static/License/License.vue
```diff
@@ -1,19 +1,102 @@
...
+      src="@/assets/fairsharing-logo.svg"
+      height="70"
+      contain
+      class="text-left pa-0 d-inline-block"
+    />
+    <p
+      :class="['mb-4 lato-font-medium lato-text-sm',{'lato-text-md':$vuetify.breakpoint.xlOnly }]"
+    >
+      The code below may be used to create a link to our logo.
+    </p>
```
knirirr, line 36:
```suggestion
    </p>
    
    <pre>&lt;img src="https://fairsharing.org/static/img/home/svg/FAIRsharing-logo.svg" alt="FAIRsharing Logo"&gt;</pre>

```

src/views/Static/License/License.vue
```diff
@@ -1,19 +1,102 @@
...
+      Except where otherwise noted, content on this site is licensed under a <a
+        href="https://creativecommons.org/licenses/by-sa/4.0/legalcode"
+        target="_blank"
+      >Creative Commons Attribution and Share-alike (CC-BY-SA) International 4.0</a> license.
+    </p>
+    <p
+      :class="['mb-4 lato-font-medium lato-text-sm',{'lato-text-md':$vuetify.breakpoint.xlOnly }]"
+    >
+      In addition, we request that any data taken from FAIRsharing, either by the API, scraping the site or other means, is attributed and cited appropriately. We request that a link and the FAIRsharing logo is presented next to the data taken from FAIRsharing.
+      In order to get full detail about how to use FAIRsharing API please visit <router-link
```
knirirr:
```suggestion
      For details of how to use the FAIRsharing API please visit <router-link
```

knirirr (approved):
This should do once the changes I've suggested below are incorporated.

PR #937, opened 2021-05-10
Organize sass file properly and got rid of css files for good

### LINKED ISSUE #908: Organizing Sass and css file
Sass file has become so long and css files are imported in main.js too much so those should be organized and chunk down into smaller files..

hoseinmirian:
The new structuring of sass files helps us to keep track of different classes and variable in a separate file. 
List of improvements:
- updated boilerplate to the latest version (8) and imported all in a sass file name "_boilerplate.scss"
- separated every type of css in a specific file: for e.g: typography holds font and font class, utilities hold margin, padding, and so on. 
- all files will be imported in an index.scss file and from now on sass files help us to organize our global class in a standard way. 
- please note the order of loading sass file in this practice matter. 
image


knirirr (changes requested):
It's useful to refactor this. However, the refactoring has caused a couple of display element changes - possibly more, but I have only noticed two. One is the login button:

Before:

Screenshot 2021-05-10 at 14 03 12

After:

Screenshot 2021-05-10 at 14 03 33

Also the dotted line on the record display cards. Before:

Screenshot 2021-05-10 at 14 05 29

After:

Screenshot 2021-05-10 at 14 05 54

It might be worth having another look yourself (I suggest comparing directly with preview in two adjacent tabs) to see if you can spot any others.

hoseinmirian:
@knirirr thanks for your review. I have applied the changes and did not spot any inconsistencies anymore.

knirirr (approved):

PR #941, opened 2021-05-12
Hossein additional info record #926

### LINKED ISSUE #926: Remaining record page fields
The fields which are handled by the additional information tab on the editing page (e.g. database-related fields) are currently not represented on the record page and ought to be added.

hoseinmirian:
added all additional information tab fields in general info provided is available. 

test record:


knirirr (changes requested):
Please put the "additional information" as a separate section - I'd suggest adding it at the bottom. 

Also, the current implementation has hard-coded the fields. The editing form uses a query to determine which additional fields are used for a particular model; see https://github.com/FAIRsharing/fairsharing.github.io/blob/master/src/store/editor.js#L225-L232

If you were to run this query and use it to determine whether or not to show additional fields, and which ones to show, then it would also resolve #942 (that should appear as an additional field).

hoseinmirian:
@knirirr all the behavior is applied based on curators feedback. If any additional fields add or delete the record display react accordingly. for test you can edit any record additional information from editing system.

knirirr (changes requested):
Currently, the additional info. section displays for collections.
But, there's only ever one additional field that applies to collections (reference URL) and due to https://github.com/FAIRsharing/fairsharing.github.io/issues/948 it will have to go into the general information section. So, this section will have to not be displayed for collections.

knirirr (approved):

Issue #947, opened 2021-05-18
More changes to related records/in collections


knirirr:
Here's a discussion:

https://pyroeis.oerc.ox.ac.uk/#narrow/stream/8-FAIRsharing-Rebuild/topic/Related.20records/near/29310

- [x] Remove third tab from in collections/in recommendations.
- [x] Change related records as described below.

The necessary change to related records will be to include all recordAssociations and reverseRecordAssociations, _unless_ it is a reverseRecordAssociation with a label of "recommends"or "collects".

If it is a recordAssociation and the linked_record is a Standard or Database it should appear in the Standard or Database tab respectively.

If it is a reverseRecordAssociation and the fairsharing_record is a Standard or Database it should appear in the Standard or Database tab respectively.

Otherwise, it should appear in the other related records tab.

PR #954, opened 2021-05-20
Hossein change in collection tab #947

hoseinmirian:


knirirr (changes requested):
Please note the following screenshot:

Screenshot 2021-05-21 at 10 21 06

The same associations have turned up in both "in recommendations" and "other related records". Furthermore, there's a problem with the latter in that these recommendations are in fact reverseRecordAssociations but the descriptive text is displayed as recordAssociations. 

For example, on the left:

The Royal Society etc. recommends Genbank  - Correct.

This same text appears on the right but is displayed as:

Genbank recommends the Royal Society etc. - Incorrect. 

Given that reverseRecordAssociations and recordAssociations are to be mixed together in the related records box it is essential that the labels for the former are in the correct order. Also, reverseRecordAssociations which collect or recommend the current record should only be displayed in "in collections/recommendations".

knirirr:
For development purposes I've added this shares_code relationship between the DNA Data bank of Japan and Genbank, so that you'll be able to check the output:

Screenshot 2021-05-21 at 10 46 54

It should appear as "The DNA Data Bank of Japan shares code with Genbank" as it's a reverse relationship (i.e. I created it on the DNA data bank of Japan page).

hoseinmirian:
@knirirr thanks for the comments. I have updated the branch, record example: Genbank
record to test:
http://localhost:8080/#/FAIRsharing.9kahy4
http://localhost:8080/#/FAIRsharing.k337f0

image


image
DNA Data Bank of Japan
(DDBJ)
image

knirirr (approved):

PR #983, opened 2021-05-25
Changed the name of a couple of labels.

knirirr:
This updates the labels for the change in https://github.com/FAIRsharing/FAIRsharing-API/issues/285


ramGranell (approved):
Adding only two words!!!

PR #1001, opened 2021-05-28
Banners

### LINKED ISSUE #670: Approval banner

### LINKED ISSUE #832: Watch record
The button needs to be connected up to the API once it is modified as here:
https://github.com/FAIRsharing/FAIRsharing-API/issues/261

In addition, something will have to be shown on the record page to indicate to users that they are watching it (e.g. a banner, icon etc.).

knirirr:
This adds banners when:

1. The record isn't yet approved by curators.
2. A user is watching the record. 

These were previously overlooked in tickets #670 and #832. 

I have marked FAIRsharing.hsbpf9 as unapproved, and you could follow this record to see the other banner.


hoseinmirian:
it looks very good even with 3 banners on top
image

hoseinmirian (approved):

PR #1023, opened 2021-06-04
Hide support section when no data available

### LINKED ISSUE #1019: Support section is not hiding
when no data available support section is not hiding
image
http://localhost:8080/#/3526

hoseinmirian:
now support section hides when no data available. 

record to test:
 http://localhost:8080/#/3526

### DIFF
--- a/src/components/Records/Record/Support.vue
+++ b/src/components/Records/Record/Support.vue
@@ -1,15 +1,13 @@
 <template xmlns:v-slot="http://www.w3.org/1999/XSL/Transform">
   <v-card
+    v-if="getField('metadata')['contacts'] && getField('metadata')['support_links']"
     class="pa-4 d-flex flex-column"
     outlined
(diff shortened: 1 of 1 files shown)


src/components/Records/Record/Support.vue
```diff
@@ -1,15 +1,13 @@
 <template xmlns:v-slot="http://www.w3.org/1999/XSL/Transform">
   <v-card
+    v-if="getField('metadata')['contacts'] && getField('metadata')['support_links']"
```
knirirr: Should this be an `||` rather than an `&&` here, e.g. if a record has support_links but no contacts, or vice-versa?

hoseinmirian: @knirirr you are right. thanks. I ll correct that.

knirirr (approved):

PR #1030, opened 2021-06-07
Add schema icon

### LINKED ISSUE #1029: define all the possible icons for support.
currently, most of the icons in support are included in the project but there are more needed to be defined.

hoseinmirian:
due to a bug that does not allow the schema_file to be saved in editor dashboard, the app is not able to show that in support and a ticket is created for that but the icon is tested by me and definition is correct in the icon files. after this ticket is merged we will be left with no icon to include for support. 

ticket for the spotted bug:
https://github.com/FAIRsharing/fairsharing.github.io/issues/1031

icon for schema_file will be this:

image

### DIFF
--- a/src/plugins/icons.js
+++ b/src/plugins/icons.js
@@ -224,6 +224,10 @@ const customIcons  = {
         home_policies: {
             type: "img",
             icon: "assets/Home/BlockInfo/home-policies.svg"
+        },
+        schema_file: {
+            type: "icon",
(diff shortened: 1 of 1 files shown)


knirirr:
We won't need this icon after all as the curators confirm they plan to handle support links differently.
I've removed the option in https://github.com/FAIRsharing/fairsharing.github.io/pull/1034.

PR #1048, opened 2021-06-09
Added countries list #917.

### LINKED ISSUE #917: Add country display to organisation page
This page: https://preview.fairsharing.org/#/organisations/1
...should also show countries.

knirirr:
This is not particularly exciting but countries are now shown on the record page.
Record 8 has several.
Record 1 has 1.


ramGranell (commented):
When there is one more than one country as organisation 8 (https://deploy-preview-1048--fairsharing.netlify.app/#/organisations/8), I will recommend to display them in the same line with comma separator (the comma is already there) instead that a different line for each country. We display the countries of records in the same line and I will keep the same format.

hoseinmirian (approved):
@knirirr @ramGranell by the new changes applied in my hotfix branch the ul structure improved and also Ramon's comment applied.

ramGranell (commented):
Only add and a space after the comma when listing the spaces as the string "Australia,Brazil,China,New Zealand,United States of America,European Union" in https://deploy-preview-1048--fairsharing.netlify.app/#/organisations/8 is not very nice

knirirr:
That's the only way I can think of to make it work.

ramGranell (approved):
Now, the list of countries looks nice.

PR #1077, opened 2021-06-18
Improve html code of duplicate-graph-link branch

hoseinmirian:
- Improve the html structure
- Lessen the logic and pushed it into HTML code so we have less code to test
- Removed test. 
- Reposition the graph button.


src/components/Records/Record/GeneralInfo.vue
```diff
@@ -38,21 +38,21 @@
...
-          class="ma-5"
-          color="primary"
-          v-bind="attrs"
-          v-on="on"
-          @click="showGraph()"
-        >
-          <v-icon
-            small
-            left
+        <router-link :to="`/graph/ ${currentRecord['fairsharingRecord'].id}`">
```
knirirr:
```suggestion
        <router-link :to="`/graph/${currentRecord['fairsharingRecord'].id}`">
```

PR #1082, opened 2021-06-21
Added copyright message.

knirirr:
This message is the same as for the client, as recommended to us.


allysonlister:
Just a quick question please. The FAIRsharing legacy website is completely under CC BY-SA 4. Doesn't that mean that our logos are already under that licence? How would that conflict (if at all) with the last line of this copyright statement?

Also, to clarify things for me please: the presence of this copyright just states that this work is ours, right? It won't conflict if we add some variant of an open-source licence later on?

Thanks!

allysonlister (approved):

PR #1094, opened 2021-06-22
Sanitised JSONLD on client.

knirirr:
This should deal with a warning for unsanitised input when merging from dev to master.


ramGranell (approved):
If you have checked that the warning does not appear when merging (it is not possible to see if I don't merge it), then go forward!!!

Issue #1099, opened 2021-06-23
Updating https://fairsharing.org/article/ redirections to new search URLs


allysonlister:
Good morning!

We have a number of redirections set up to aid readability for the searches used in our paper (https://doi.org/10.1038/s41587-019-0080-8). Please could these be updated to new search URLs when the rebuild goes live? Also, please could someone double check that the new URLs I've suggested look right for the rebuild? I'm guessing, for instance, that the "#" will need removing?

1. https://fairsharing.org/article/live_list_standards_in_policies  
- old URL: https://fairsharing.org/standards/?q=&selected_facets=recommended:true
- new URL: https://dev--fairsharing.netlify.app/#/search?fairsharingRegistry=Standard&isRecommended=true&page=1
2. https://fairsharing.org/article/live_list_databases_in_policies
- old URL: https://fairsharing.org/article/live_list_databases_in_policies
- new URL: https://dev--fairsharing.netlify.app/#/search?fairsharingRegistry=Database&isRecommended=true&page=1
3. https://fairsharing.org/article/live_list_journal_policies
- old URL: https://fairsharing.org/article/live_list_journal_policies
- new URL: https://dev--fairsharing.netlify.app/#/search?fairsharingRegistry=Policy&recordType=journal&page=1

Thanks!

knirirr:
These are defined here in the old system, as well as another redirection which might usefully be dealt with at the same time:

https://github.com/ISA-tools/biosharing/blob/master/biosharing_web/views/general.py#L1728-L1751

PR #1106, opened 2021-06-28
Redirections 1099

knirirr:
Ticket #1099 mentions various URLs to be redirected internally, and my comment there mentions some external ones (/ontologies/DRAO and /ontologies/SRAO). These should redirect as specified.


ramGranell (approved):
The URLs of the tickets seem to work properly

Issue #1105, opened 2021-06-25
Create a link to user public profile in editor


hoseinmirian:
create a link representing the user profile and have a copy button next to it, being able to copy the URL to share, etc. 

sample:
https://fairsharing.org/users/659

PR #1108, opened 2021-06-28
Added link as per #1105.

knirirr:
Now, a link should be visible to your public user profile (/users/:id) on your private profile (/accounts/profile). 
Please feel free to create a PR to this one with your preferred appearance if you don't fancy the looks.


ramGranell (approved):

hoseinmirian (approved):
I think it looks fine for now but later in a UI/UX examination and improvement after MVP we can better the whole dashboard.

PR #1117, opened 2021-07-05
Organisations and users 298

knirirr:
This PR shows users on any organisation page.
For example, /organisations/2274. 
It should not show the user's email address if they have requested to hide it.


ramGranell (approved):
That's great! Just one comment, when the fields are empty e.g. "Alternative Names: None" maybe there is not need to add it. Anyway, we are doing the same for many record fields.

knirirr:
> That's great! Just one comment, when the fields are empty e.g. "Alternative Names: None" maybe there is not need to add it. Anyway, we are doing the same for many record fields.

I suppose I could remove it for consistency with the other fields.

PR #1143, opened 2021-07-12
History permissions for curators etc.

### LINKED ISSUE #1141: Auth for logs
Currently, record history is visible only with the privileges of a normal user.
Authentication should be sent when getting history, for the benefit of curators.

knirirr:
https://github.com/FAIRsharing/FAIRsharing-API/issues/305 includes some modifications to the data returned for record history. As part of this I've modified the client so that only curators can see changes to curator notes. 

To try this out:
1. Edit a record as a curator, changing both description and curator_notes.
2. View history as a normal user, only description changes show.
3. View as a curator, curator_notes changes now also visible.


ramGranell (approved):
As far as I can see (as a curator and as accessing without being a user) it works properly.

hoseinmirian (approved):

PR #1149, opened 2021-07-13
Added some text for #1123.

### LINKED ISSUE #1123: Placeholder text for empty ORCIDs in user profiles
Please could you some text to the user profile edit form (and perhaps just the user profile page generally as a placeholder if no ORCID is listed by the user) stating that providing an ORCID is strongly encouraged.

Here is the text: "We strongly recommend including your ORCID to provide extra information for the FAIRsharing community about you and the resources you develop."

Thanks!

knirirr:
If a user goes to edit their own profile via the usual link (/profiles/edit) and hasn't set their ORCID ID they should be shown a message at the bottom of the edit form encouraging them to do so.

### DIFF
--- a/src/views/Users/EditProfile.vue
+++ b/src/views/Users/EditProfile.vue
@@ -95,6 +95,13 @@
                     get in touch</a>.
                 </p>
 
+                <p
+                  :v-if="!user().metadata.orcid"
+                >
+                  <b>We strongly recommend including your ORCID ID to provide extra information for the FAIRsharing
+                    community about you and the resources you develop.</b>
+                </p>
+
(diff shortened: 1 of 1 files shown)


ramGranell (commented):
When editing my profile, the text "We strongly recommend including your ORCID ID to provide extra information for the FAIRsharing community about you and the resources you develop." appears always independently I have or I haven't the ORCID number, is this the expected behaviour?

knirirr:
The expected behaviour is that if you load the edit profile page without having an ORCID associated with your account then you will see the message.
What you do once the page is loaded will have no effect.

ramGranell (approved):

PR #1152, opened 2021-07-15
Don't show review banner when logged out.

knirirr:
This fixes a small bug which was causing the curator review banner to be displayed when logged out.


ramGranell (approved):
Yep, it seems to work properly

PR #1153, opened 2021-07-15
Hossein link userorganisation #1098

### LINKED ISSUE #1098: Linking users to organisations
Please see: https://github.com/FAIRsharing/FAIRsharing-API/issues/298

Client-side, the following are necessary:

- [x] List organisations on users public and private profile pages.
- [x] Allow deleting organisations on users' profile edit pages,
- [x] Add list of affiliated users to organisation pages. 

Users should be able to add themselves to any organisation and create a new one (like user defined tags) if they can't find one they like.

hoseinmirian:
Please check OrganisationTable both in private profile and public profile
-- Please ignore the bottom table of the component due to it will be completed in the next ticket:https://github.com/FAIRsharing/fairsharing.github.io/issues/1154

what to test:
1- logged out users cannot change the record assigned to a user but only see them (delete button is hidden for them) all the time and everywhere.
2- logged in users with normal access, can delete only its own organizations either whilst they are viewing their own profile from a public profile like /users/:id or they can edit their own organizations form their private profile from accounts/profile (please note if they see other public profiles, they only are able to see the organisations attached to a user not able to delete them)
3- logged in super curator can edit users profiles from anywhere.. (they will see delete button next to each organization all the time)


knirirr (changes requested):
I noticed the following whilst trying out the application:

The organisation edit form was visible, though non-functioning, when I did not have permission to edit. 

Screenshot 2021-07-20 at 14 21 00

In any case, I'd expect to see this edit form on /profiles/edit or /profiles/editPublicProfile/:id rather than /accounts/profile; the latter page is intended for viewing only. 

It wasn't functioning (i.e. I could not search for organisations) but presumably that was intended.

hoseinmirian:
@knirirr please ignore the bottom table as I explained in my first comment. just please focus on the user organisations table. thanks

knirirr:
> @knirirr please ignore the bottom table as I explained in my first comment. just please focus on the user organisations table. thanks

No; this table is not on the correct page, Please move it to the edit page for the user - thanks.

hoseinmirian:
OrganisaitonTable component moved to public and private profile and will only be accessed for signed-in users. 
- feel free to assign organisation from available organisations to user organisations table
- Data will instantly update in tables
- Create a new organisation and instantly it will be assigned to the current user (you can do it from public or private profile for testing)
- Removing a single organisation from the user's organisation table.

knirirr (changes requested):
As far as I can tell the functionality for creating organisations and (dis-)associating them with users is as it should be. But, there are still some layout issues that will need addressing. Most notably, the organisation tables have disappeared from the private and public profile pages. This means that a user can't see their own organisations until they start to edit, and only a curator may see anyone else's. Here's the private profile before and after:

private_old
private_new

...and the public:

public_old
public_new

In this latter the maintained records table has changed position and size, probably because of the missing organisation table.

When adding a new organisation any change to the organisation type causes it to expand horizontally. This ought to be addressed as it may cause problems on small screens or be disconcerting:

expanding

Finally, when editing another user's profile as an admin, the effect of the update button shown below is only to show an update message. It would be better to exit back to the user's display page (this is what happens when clicking the button on one's own profile):

update_button

hoseinmirian:
in regard to the bug with autocomplete adding items I investigated and on small screen there is no concerns but in large and xl screens I restricted the max-width to 1200px so if more items get added into it, it will only grow up to 1200px not more..

small screen is safe. 
image

knirirr (changes requested):
Here are images from of the public and private profiles:

public_profile

private_profile

The areas outlined in red (empty space, extraneous edit forms on a viewing page) will need to be removed without modifying the edit pages.

knirirr:
As everything's already in #1176, including the Codacy fix, I'll close this PR and merge #1176 to dev.

PR #1171, opened 2021-07-22
Improve the claim now in maintainer

### LINKED ISSUE #1169: Updated text for "This record is in need of a maintainer" in record display
Thank you for having different text depending on if a user is logged in or not. I'd like to modify the text a little, as we want to advertise that the record can be claimed, even if people haven't logged in yet.
claim

- [ ] If logged out, the text should say "This record is in need of a maintainer. If you are affiliated with this project, log in and claim it now!" (with "log in" being the text that is hyperlinked).
- [ ] If logged in but there is no maintainer yet, say "This record is in need of a maintainer. If you are affiliated with this project, claim it now!" (with "claim it now" hyperlinked). This is the same text as it is used now, but with a lowercase "c" as it is the middle of a sentence.

Thanks!

hoseinmirian:
please be logged out and try a record that needs claiming as well as being logged in and check for a record that needs maintaining...


src/components/Records/Record/GeneralInfo/Maintainers.vue
```diff
@@ -6,21 +6,29 @@
...
-          class="ma-0 mr-2"
+          v-if="!canClaim && !user().isLoggedIn"
+          class="ma-0 mr-1"
         >
-          If you are affiliated with this project,
+          log in and
         </p>
+        <router-link
+          v-if="!canClaim && !user().isLoggedIn"
+          to="/login"
```
knirirr: This should also be `@click="()=>{$emit('requestOwnership')}"` (as is "claim it now"), rather than a link to /login.

knirirr (changes requested):
This is fine other than the link to /login, which is not a valid URL. This did, however, reveal another bug I must deal with myself.

src/components/Records/Record/GeneralInfo/Maintainers.vue
hoseinmirian: image

Allyson also in the comment mentioned turning login into a link.. please check if it's correct. link to login is correct now. tnx

knirirr (changes requested):
The login link is not what I suggested, but it will do.
After logging in and requesting a to maintain a record by clicking on "claim it now" the text disappears, resulting in this odd-looking result:

Screenshot 2021-07-26 at 10 37 36

Disabling the link but leaving it present if canClaim() is false (as it should be if a pending claim request is present) would avoid this. After modifying it, please check the behaviour after clicking on this link.

hoseinmirian:
what happens currently in dev branch is that after requesting to maintain while being logged in, claim now link disappears..

or after claiming you like the text to be like below?
image

src/components/Records/Record/GeneralInfo/Maintainers.vue
```diff
@@ -22,7 +27,13 @@
...
           class="ma-0 mr-1"
         >
-          and claim it now!
+          and
+        </p>
+        <p
+          v-if="!canClaim"
+          class="ma-0 mr-1"
+        >
+          claim it now!
```
knirirr: Probably better to remove this phrase altogether if this is the way you'd rather do it. So, instead of:

`This record is in need of a maintainer. claim it now!` (note dodgy capitalisation)

...it would be:

`This record is in need of a maintainer.`

knirirr (changes requested):
This appears to work other than the formatting issue mentioned separately.

knirirr (approved):

PR #1180, opened 2021-07-26
Delete useless viewOrganisation and replaced it with OrganisationTable

hoseinmirian:
I deleted viewOrganisation component due to organisationTable was already customized to show or not show modifing tools based on the current route. Besides it has a searchbar which is handy for users with multiple organisations


knirirr (changes requested):
The ViewOrganisations (essentially the re-named original Organisations table) serves to form a simple means to display only organisations; the modified OrganisationsTable is now rather complex and, when displayed alongside the other tables on the public and profile tables, looks different.

hoseinmirian:
@knirirr ok then I ll close this PR

PR #1181, opened 2021-07-27
Record review 1150

knirirr:
This adds an additional button to the curator dashboard which supplies a file listing all records which haven't been reviewed by a curator within the last year. 
None of the records will have a previous review date yet as none have ever been reviewed.

### DIFF
--- a/src/lib/GraphClient/queries/curators/getSummary.json
+++ b/src/lib/GraphClient/queries/curators/getSummary.json
@@ -2,6 +2,7 @@
   "queryName": "curationSummary",
   "queryParam": null,
   "fields": [
+    "needsReview",
     {
       "name": "curatorList",
       "fields": [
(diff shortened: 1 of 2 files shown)


hoseinmirian (approved):
@knirirr  branch looks fine. I appreciate very shortly explain about this line. tnx


          this.downloadReviewContent = "data:text/json;charset=utf-8," + encodeURIComponent(this.allDataCuration.needsReview.join('\n'));

knirirr:
Certainly.
`downloadReviewContent` is an array of lines of data which which needs to be converted to a suitable string format and streamed as a file. So, adding `"data:text/json;charset=utf-8,"` at the beginning and encoding it informs the browser that it's to be treated as an attachment. 
There's a similar technique used for one more file on the page.

PR #1187, opened 2021-07-29
Search hackery

### LINKED ISSUE #1186: Outdated URLs
Though some search URLs from the old system have been dealt with correctly (aliases implemented in the new as well), others are hanging around (e.g. in papers) which cannot be parsed by the new system. Examples:

```
https://fairsharing.org/search/?q=genomics&content=biodbcore
https://fairsharing.org/search/?q=transcriptomics&content=biodbcore
https://fairsharing.org/search/?q=genomics&content=standard&name=&taxonomies=&organisations=&shortname=&description=&supportlinks=&licenses=&countries=&maintainers=&expanded_onto_domains=&expanded_onto_disciplines=&user_defined_tags=&record_id=&miriam_id=&search_state=hidden
https://fairsharing.org/search/?q=proteomics&content=biodbcore&name=&taxonomies=&organisations=&shortname=&description=&supportlinks=&licenses=&countries=&maintainers=&expanded_onto_domains=&expanded_onto_disciplines=&user_defined_tags=&record_id=&miriam_id=&search_state=hidden
https://fairsharing.org/search/?q=proteomics&content=standard&name=&taxonomies=&organisations=&shortname=&description=&supportlinks=&licenses=&countries=&maintainers=&expanded_onto_domains=&expanded_onto_disciplines=&user_defined_tags=&record_id=&miriam_id=&search_state=hidden
https://fairsharing.org/search/?q=metabolomics&content=biodbcore&name=&taxonomies=&organisations=&shortname=&description=&supportlinks=&licenses=&countries=&maintainers=&expanded_onto_domains=&expanded_onto_disciplines=&user_defined_tags=&record_id=&miriam_id=&search_state=hidden
https://fairsharing.org/search/?q=metabolomics&content=standard&name=&taxonomies=&organisations=&shortname=&description=&supportlinks=&licenses=&countries=&maintainers=&expanded_onto_domains=&expanded_onto_disciplines=&user_defined_tags=&record_id=&miriam_id=&search_state=hidden
```

Some means of handling these which minimises as much as possible the impact on maintainability and speed would be useful, if possible.

knirirr:
This code causes search parameters to be modified if they match certain (known) patterns from the old system. 

This ticket lists some particular URLs which have been used in a paper recently and so must be intercepted, as readers of the paper will encounter the new site and try to use old URLs with it.

https://github.com/FAIRsharing/fairsharing.github.io/issues/1186

Look in `src/router/hackSearch.js`  for details of the re-writing. It's called any time anyone visits a search url (see `src/router/index.js`). The `modified` parameter should ensure that the hack is only called once. 

Of course, queries formatted for the new system should pass through `hackSearch.js` untouched.


hoseinmirian (changes requested):
Thanks, it works fine very good but there is a minor issue. For the first visit or after refreshing the page whilst having an invalid query string it works fine but after the page is loaded and we press back on the browser it will include and accept the query as a valid query in the chips holder, please check both URL and also chips holder situation:

image

knirirr:
I'm not sure that this is something that can be prevented; this whole PR is a bit of a hack in any case.
Do you have any suggestions as to how to accomplish this? If there happened to be some way to call `next()` without the page with incorrect params ending up in history it would be useful, but I don't think there is here:

https://github.com/FAIRsharing/fairsharing.github.io/blob/9ae13c5df1b5eb9efac349b8019de7a80cc6f634/src/router/index.js#L96-L98

hoseinmirian:
> I'm not sure that this is something that can be prevented; this whole PR is a bit of a hack in any case.
> Do you have any suggestions as to how to accomplish this? If there happened to be some way to call `next()` without the page with incorrect params ending up in history it would be useful, but I don't think there is here:
> 
> https://github.com/FAIRsharing/fairsharing.github.io/blob/9ae13c5df1b5eb9efac349b8019de7a80cc6f634/src/router/index.js#L96-L98

I tried some code today to see how I can fix it. still no luck

hoseinmirian:
@knirirr 

My Final review

I think no user can add those edge-cases filters into the URL by using our UI, and the only way they might be redirected would be an old link given to them or shared with them(which prevent them your code properly in these cases), then the code you already had written I believe is enough and its very rare that someone enters a new filter(query) in the URL or press back. Besides we are not responsible for all sort of data user enters in the query if no data is returned then there will be no problem in this regard as well. 

Due to the above reasons I approve your branch.

hoseinmirian (approved):

knirirr:
Thanks, sounds reasonable.
I'll not merge this yet, in case anyone else wants to comment. In the absence of any further suggestions I'll do so before we switch to the new system so that those old links will be caught.

PR #1192, opened 2021-08-13
Hossein change history mode #805

### LINKED ISSUE #804: Internal links
Some internal links are constructed incorrectly, for example:

https://github.com/FAIRsharing/fairsharing.github.io/blob/master/src/components/Records/Record/Maintainers.vue#L53

This is `<a :href="maintainer.username + '/' + maintainer.id">` and should be `<a :href="'/users/' + maintainer.id">`

But, even then, this link will fail as it will go to /users/1#/ rather than /#/users/1. The question is therefore if it's necessary to put the hash in every internal URL, or if some function exist that will add this depending on the status of history mode.

See also: https://github.com/FAIRsharing/fairsharing.github.io/pull/786#discussion_r595901605

### LINKED ISSUE #805: Changing reference all links (href) to history mode
We will need to go history mode in production, which means that we will need to remove all href '# in the reference links, e.g. 
 <a :href="'#/' + props.item.id"> needs to be <a :href="props.item.id"> in https://github.com/FAIRsharing/fairsharing.github.io/blob/dev/src/views/Curators/Curator.vue#L68

hoseinmirian:


hoseinmirian:
@knirirr I activated history mode so we got rid of hashtag. everything seems normal in local but in the github deploy there is a minor issue which I believe its related to server configuration. So lets say in the Educational route, if we refresh the page while having a URL containing anchors like educational/#faq1-2, the page does not refresh properly and get an error not found page.

knirirr:
So far I've noticed that links on the curator dashboard result in a 404, e.g. https://deploy-preview-1192--fairsharing.netlify.app/3565
Presumably it would be fixed if we could deploy this on a server with a properly configured Apache installation, or if the links on the curator dashboard were re-written as mentioned in #805 . I'm still looking into this, though.

knirirr:
I plan to delay merging this one until I've tested the branch on a suitably-configured server.

knirirr (approved):
I'm happy with this (I've made some changes) but don't fancy merging it until I have some feedback from use of the beta site.

allysonlister:
I'm not sure if I add review comments here, so please redirect me if they belong elsewhere...

The record navigation and searching seems to work well, thank you. 

I am also echoing the bug that @hoseinmirian noticed with regards to the URLs for the FAQ items in Educational.

1. visit https://beta.fairsharing.org/educational
2. click on faq item 3.3
3. the webpage directs you to https://beta.fairsharing.org/educational#/educational#faq3-3 (note the odd # and I think the URL isn't right)
4. instead of opening up / expanding 3.3 upon user click, it instead jumps you to the top of the page. If you scroll down to 3.3 a second time, you are now able to expand/collapse item 3.3

I have done some very basic editing as well, and it seems to be working so far. With this in mind, would you like me to approve now, or do further tests?

Thanks very much!

knirirr:
@allysonlister if you have time to try more of it then please do.
@hoseinmirian please would you deal with the educational page bug (which will include fixing the home page link). 
You'll need to set VUE_APP_MODE to 'history' in your .env.

hoseinmirian:
> @allysonlister if you have time to try more of it then please do.
> @hoseinmirian please would you deal with the educational page bug (which will include fixing the home page link).
> You'll need to set VUE_APP_MODE to 'history' in your .env.

ok thanks I work on it.

PR #1198, opened 2021-08-19
Hossein display deprecated #948

### LINKED ISSUE #948: Displaying deprecation records
See https://docs.google.com/document/d/1K7i_mmV5syR9ZDvUvg4kDROCvUN4kWWIEtMfsoBWz7Y/edit, though we cannot do exactly what is described there.

> Add “Deprecation Information” in red after the DOI / before “Type”. This consists of (Example: https://fairsharing.org/FAIRsharing.7xkx69):
> Deprecation Note would read: “This record was deprecated on April 10, 2018 for the following reason(s): This resource is now obsolete and, together with CSA (https://fairsharing.org/FAIRsharing.2ajtcf), has formed the basis of the M-CSA resource.”
> Please note this incorporates the “deprecation date” field.
> If a “deprecates” relationship exists in the incoming or outgoing direction, please list those relationships. In this example it is an incoming relationship and it would say “This record is replaced by: Mechanism and Catalytic Site Atlas“. If there were an outgoing “deprecates” relationship, another sentence would be added e.g. “This record replaces or incorporates the following deprecated resources: XYZ”

This is dependent on the completion of https://github.com/FAIRsharing/FAIRsharing-API/issues/282 first.

hoseinmirian:
Please test this record and click on its replaced deprecated link to see the other record with different data
http://localhost:8080/#/FAIRsharing.7xkx69


src/components/Records/Record/GeneralInfo/ReplacedByRecord.vue
```diff
@@ -0,0 +1,58 @@
+<template>
+  <div
+    v-if="(!!currentRecord['fairsharingRecord'].reverseRecordAssociations.find(item=>item.recordAssocLabel==='deprecates') && currentRecord['fairsharingRecord'].reverseRecordAssociations) || (!!currentRecord['fairsharingRecord'].recordAssociations.find(item=>item.recordAssocLabel==='deprecates' && currentRecord['fairsharingRecord'].recordAssociations))"
```
knirirr: This if statement is fairly complex and could be hard to understand when it comes to making changes to this code later.
Though it will mean adding an extra test, it would be worthwhile abstracting this calculation into a separate method. The method could then be commented to describe what it does, e.g.

```
<div
    v-if="replacedOrReplaces"
    ...
>
```

```
methods: {
    // Return true if this record deprecates another or is deprecated by another. 
    replacedOrReplaces: () => {
        // find relations in (reverse) associations here and return a boolean.
    }
}
```

The same could possibly be applied to other v-ifs here, but this one is the most important as I think it will be the hardest for future maintainers to deal with.

knirirr (changes requested):

knirirr (changes requested):
Though there are no comments, the separate functions with clear names should make this much easier to deal with.
Unfortunately there seems to have been a regression in the tests at some point. For example:

https://github.com/FAIRsharing/fairsharing.github.io/pull/1198/checks?check_run_id=3381659032

```
PASS tests/unit/store/users.spec.js

  ●  Cannot log after tests are done. Did you forget to wait for something async in your test?
    Attempted to log "[Vue warn]: Error in mounted hook (Promise/async): "TypeError: Cannot read property 'querySelector' of null"

    found in

    ---> <Record>
           <Root>".

      at console.error (node_modules/@jest/console/build/BufferedConsole.js:163:10)
      at warn (node_modules/vue/dist/vue.runtime.common.dev.js:621:15)
      at logError (node_modules/vue/dist/vue.runtime.common.dev.js:1889:5)
      at globalHandleError (node_modules/vue/dist/vue.runtime.common.dev.js:1884:3)
      at handleError (node_modules/vue/dist/vue.runtime.common.dev.js:1844:5)
      at node_modules/vue/dist/vue.runtime.common.dev.js:1861:39
```

I recall you dealing with some of these warnings before; if the same methods can be employed here to get rid of them then this PR should be fine to merge.

knirirr (approved):
On second thoughts, please proceed with this and I'll open a separate bug ticket for the tests.

PR #1199, opened 2021-08-20
Detached pagination of being coupled with records currentPage and rep…

hoseinmirian:
Pagination was not fully decoupled and was relying on the currentPage from records store. I replaced it with props and passed it from its parent ListController.vue which helps to keep it more reusable.


src/components/Records/Search/Header/ListController.vue
```diff
@@ -137,7 +140,7 @@
             }
         },
         computed: {
-            ...mapState('records', ["totalPages", "loading"]),
+            ...mapState('records', ["totalPages", "loading","currentPage"]),
```
knirirr:
```suggestion
            ...mapState('records', ["totalPages", "loading", "currentPage"]),
```

knirirr (approved):

PR #1207, opened 2021-08-31
Graph modifications 1163

### LINKED ISSUE #1163: Front-end graph improvements
After discussions with @knirirr and @terazus we are requesting some updates to the relationship graphs please.

- [ ] **Node update: please change node shapes to be smaller versions of the type icons.** In the past, @terazus thinks that there was a ticket concerning the replacement of the existing node shapes with small images for each record type (database, standard, policy, collection) based on the type icons. This may need @hoseinmirian to redraw these icons smaller. In particular, please can we ensure that each type icon is a different colour so that they stand out in both colour and design. You can look at the colours used in the legacy graphs by visiting any collection's graph page, e.g. https://fairsharing.org/graph/#/collection/bsg-c000090 (you may need to modify the colours slightly to fit with the rebuild's colour scheme).
- [x] **Colour for relationships**: If possible, colour relationships differently and move the names of the relationships to the legend.
- [ ] **Count number of nodes**, and if nodes are too high (e.g. >50) then start with level 1 rather than level 2 to help our users with slower machines :-)
- [x] **Type toggles**: be able to show/hide record types on the graph. 

Please let me know if you have any questions, and thanks!

knirirr:
This branch makes graphs a single level, but much quicker as the data are generated on the server.
In order to work it will need the changes from https://github.com/FAIRsharing/FAIRsharing-API/pull/315 plus the rake task running to regenerate the graphs.


knirirr:
@terazus the relevant branch is now running on the server so it should be possible to try this out.

hoseinmirian (approved):
everything seems fast and normal. tnx

PR #1215, opened 2021-09-06
Hossein display user #1204

### LINKED ISSUE #1204: Display messages to the user
Messages are set on the server to warn users of such things as maintenance or other disruption.
There's a ticket describing how the client should handle them here:

https://github.com/FAIRsharing/FAIRsharing-API/issues/301#issuecomment-905475905

This ticket is simply for the banner display (and getting the data to display). Tools for curators to modify messages can come later.

hoseinmirian:


hoseinmirian:
publicMessage compoennt shows all the public announcements defined in the back-end.

knirirr (changes requested):
It does the trick in that the banners display and a page refresh is required if there's any change in messages on the server (which is what I was after). 

But the code structure is not what I had in mind in that it's using src/store/searchFilters.js to get the messages. For maintainability, it would be better to use src/store/messages.js. 

For the appearance of the messages the dates will need formatting as has been done on record pages, and a bit of a margin either side of the banner wouldn't go amiss either.

Screenshot 2021-09-06 at 15 55 14

hoseinmirian:
thanks for the review. One question: what do you mean by the screenshot by saying border? did you mean margin?

knirirr:
> thanks for the review. One question: what do you mean by the screenshot by saying border? did you mean margin?

I mean a bit of white space where the arrow's pointing (and at the same place on the other side). Apologies for the unclear label; had some trouble with Preview.

hoseinmirian:
> > thanks for the review. One question: what do you mean by the screenshot by saying border? did you mean margin?
> 
> I mean a bit of white space where the arrow's pointing (and at the same place on the other side). Apologies for the unclear label; had some trouble with Preview.

ok thanks. I ll apply your changes.

hoseinmirian:
applied comments
image

knirirr (approved):

PR #1221, opened 2021-09-09
Make icons smaller in record display

hoseinmirian:


knirirr (approved):
This all appears to work, though I don't think it offers any improvement in appearance, unfortunately.

PR #1227, opened 2021-09-14
Hossein network graph new #835

### LINKED ISSUE #835: Network graph icons
Currently, the network graph shows geometric shapes to indicate the record registry. instead, icons could be shown, but special icons of a smaller size (15px?) would be required. But, for now, the current shapes are sufficient. 

This would probably also require modifications to the API to indicate the required icon. 

our post in hichart forum:
https://www.highcharts.com/forum/viewtopic.php?f=9&t=47527&p=171944#p171944

created github ticket of the issue
https://github.com/highcharts/highcharts/issues/16659

hoseinmirian:
NetworkGraph component got updated with new svg icons


knirirr (changes requested):
Good to see that the Github warnings have been dealt with - nice catch.
However, there are various issues with the UI that will need to be dealt with:

1. It's not clear which is the central record; clicking on this has a different result from clicking on any other and it ought to be marked separately as on the previous version. For example, by changing the text to red.

Screenshot 2021-09-15 at 13 58 01

2. The switches which allow selecting of different registries have ceased to function. Compare with the previous version for the same graph:

Screenshot 2021-09-15 at 13 58 25
Screenshot 2021-09-15 at 13 58 52

3. Some standards do not have any icon associated with them. 

Screenshot 2021-09-15 at 14 04 10

4. There ought also to be a list of icons and types/registries below the switches to make particularly clear which is which on the graph. Compare the graph in the old system:

Screenshot 2021-09-15 at 14 06 49

With all these done is should be OK for deployment.

knirirr (changes requested):
Unfortunately the icons don't appear when an image is saved - this will need fixing before this can be merged.


FAIRsharing-network-graph

Also, there's something funny going on with the legends in different browsers. For example:

Firefox:
firefox

Safari:
safari

Chrome:
chrome

knirirr:
For future reference: https://www.highcharts.com/forum/viewtopic.php?f=9&t=47527&sid=39755dc677a088bbdf1f27272d91f6b9

knirirr:
This can be closed in favour of changing the icon colours.

PR #1268, opened 2021-10-11
Pass parameters when approving

knirirr:
This should cause records which are approved by the super_curator on the curation dashboard to be marked as reviewed by said super_curator.

### DIFF
--- a/src/components/Curators/RecordsAwaitingApproval.vue
+++ b/src/components/Curators/RecordsAwaitingApproval.vue
@@ -413,7 +413,8 @@
               let preparedRecord = {
                 approved: true,
                 skip_approval: true,
-                processing_notes: null
+                processing_notes: null,
+                create_review: true
               };
               let data = {
                 record: preparedRecord,


ramGranell (approved):

allysonlister:
Originally, I had requested that this action (marking the record as reviewed) only happen during record _creation_, not during either record _creation or record update_ (see https://github.com/FAIRsharing/FAIRsharing-API/issues/339). Right now, the review banner is removed upon approval of either new records or updated records. If this is the only way this can be done, then I will approve the review. However, the original intent is that it was only newly-created records that this makes sense for.

knirirr:
> I had requested that this action (marking the record as reviewed) only happen during record creation

This cannot happen because there is no curator present when a user creates a record via the fairsharing_records_controller. 

>  Right now, the review banner is removed upon approval of either new records or updated records. 

Given the original requirements (https://github.com/FAIRsharing/FAIRsharing-API/issues/96), I think this is the correct behaviour. 
The purpose of the record review, according to the discussions I had at the time, was to identify records which have not been edited in > 1 year and allow curators to mark them as still correct even if those records don't require editing.
A similar objective could be achieved by marking all records that haven't been edited for > 1 year as unapproved and having a curator click an approval button; approval of a record implies that the curator has inspected it and deemed it acceptable, as does clicking the review option in the action menu on a record page. But, the additional complexity of the record_review allows the name of the reviewing curator to be associated with the record for this purpose and also avoids flagging such old records to users because only curators can see the review banner.

allysonlister:
Thanks @knirirr  - I will approve :-)

allysonlister (approved):

PR #1276, opened 2021-10-14
Aligned registry type and abbreviation in cardStack

### LINKED ISSUE #1210: Abbrevation and comments should be centered vertically
image

hoseinmirian:
now the abbreviation and registry type is aligned center vertically. please check RecordsCardStack


knirirr:
The Netlify build I'm looking at doesn't have centered abbreviations similar to what's shown in the original ticket - they're aligned to the left:

Screenshot 2021-10-14 at 15 56 12

hoseinmirian:
> The Netlify build I'm looking at doesn't have centered abbreviations similar to what's shown in the original ticket - they're aligned to the left:
> 
> Screenshot 2021-10-14 at 15 56 12

that is correct due to I did center abbreviation but it looked bizarre so I decided to keep it on left. it looks more aligned this way.

knirirr:
The requirements in the ticket state that they should be central, though:
https://github.com/FAIRsharing/fairsharing.github.io/issues/1210

hoseinmirian:
abbreviation become center in search cards. 
image

knirirr (approved):

PR #1278, opened 2021-10-15
Fixed logo disappearance issue

### LINKED ISSUE #1247: Logo disappears in an inconsistent manner from the left side of the header
**Describe the bug**
the FAIRsharing logo disappears sometimes from the top left of the pages when using the website. Often - but not always - this happens as a consequence of editing a record. It also happens at other times that are not always repeatable.

**To Reproduce**
1. Go to https://beta.fairsharing.org/FAIRsharing.9kahy4
2. Pretend to edit it (please don't actually edit it! this is a live database!)
3. Watch the logo disappear.
Sometimes this also happens when I go through the forgot password route when logging in.

**Expected behavior**
The logo should remain

**Additional context**
This is an intermittent problem and as such will be hard to diagnose. I know that Dom has sometimes seen this behaviour too.

hoseinmirian:
logo v-img has been replaced with img tag so disappearance is not happening anymore.


knirirr (approved):
Appears to work on Netlify. We can only merge this and see how it behaves in production.

PR #1281, opened 2021-10-15
Graphs without data

### LINKED ISSUE #1279: Deal with empty edges array in graphs
If a record's graph data has no edges (i.e. it's not linked to any other record) then the graph fails to display as the javascript on the page is expecting to see something in that array. This will need to be handled somehow - perhaps by displaying a "no relations" message if that array is empty. 

https://github.com/FAIRsharing/FAIRsharing-API/issues/349

knirirr:
This should display a nice message if a graph doesn't have any links or hasn't had its data generated. 

Example: https://deploy-preview-1281--fairsharing.netlify.app/#/graph/1234


terazus (changes requested):
Unfortunatly, the error windows shows up during data loading:
https://deploy-preview-1281--fairsharing.netlify.app/#/graph/3493 for instance.

I would only show that window after the API call and if there's no data or an error.

knirirr:
> Unfortunately, the error windows shows up during data loading: https://deploy-preview-1281--fairsharing.netlify.app/#/graph/3493 for instance.

Good spot. It was off the bottom of my screen so I couldn't see it initially.

terazus (approved):

PR #1296, opened 2021-10-21
Page title fix

### LINKED ISSUE #1295: Title in the browser tab doesn't update when no acronym available
**Describe the bug**
When I leave a record from another record page, then search for a new record using the header that does not have an acronym to populate the browser tab's title bar, the tab title is populated with the name of the old record and isn't updated to reflect the new page the user is on.

**To Reproduce**
1. Go to https://beta.fairsharing.org/1380
2. search using the header for a record without an acronym, e.g. "FAIR chemical structure template"
3. The search page (https://beta.fairsharing.org/search?q=FAIR%20chemical%20structure%20template) shows "FAIRsharing | search" but then when I click on the new record page it shows "FAIRsharing | NBCA-R" still (https://beta.fairsharing.org/1539)
4. This behaviour does not happen when searching for a record with an acronym. From 1539, use the search bar in the header and search for "ddbj". 
5. Click on the DDBJ record (https://beta.fairsharing.org/2894) - you'll notice that that acronym in the browser tab title is updated.

**Expected behavior**
The browser tab title to reflect the page the user is on. If no acronym, I believe the current behaviour is to show only "FAIRsharing". I would be happy to carry on doing this, or just provide the first X characters from the  full name (where X is a number that you deem appropriate)

**Screenshots**

Ubuntu / Firefox

knirirr:
I couldn't replicate the behaviour described in #1295 but I hope this change will fix it for those affected.


allysonlister:
Thank you - this branch works correctly. I can see the NBCA-R string briefly when the "FAIR chemical structure template" is loading, but it is quickly replaced by "FAIR chemical structure template". Thanks!

allysonlister (approved):

ramGranell (approved):

PR #1304, opened 2021-10-28
Duplicates 1249

### LINKED ISSUE #1249: Add check for name/abbreviation/homepage already in FAIRsharing during record creation
We had a really useful check in the legacy site that listed any records with similar name/abbreviation/homepages when a user was creating or updating a record and was editing those fields. It would be fantastic if a similar thing could be implemented in the new system, as we've already had someone create a new record, not realising that there already was one.

Thanks!

knirirr:
This needs the server-side code in https://github.com/FAIRsharing/FAIRsharing-API/pull/365 in order to function; this is currently deployed on the dev API server. 

What it should do is query the server when a user chooses to create a record and stop them if the name, abbreviation or homepage match anything found via an Elasticsearch query. They can then choose to clear the name and try again or carry on regardless. Please create a PR to this branch if you wish to change the appearance.


knirirr:
I've deployed the server changes on dev-api so this can be tried out.
Please note that if a record is created despite the warning then a processing note will be added in the curator dashboard.

hoseinmirian (approved):

ramGranell (approved):

allysonlister:
Thanks!

This is all working really well. I like the red box, and the "I know what I'm doing". I like that if I make the record anyway, a processing note is automatically added.

I've tried a few permutations, and only noticed one oddity. When using the string "miame" in either the name or abbreviation, the duplicate check only finds NuML (see screenshot below), even though there are lots of MIAME-related records (https://deploy-preview-1304--fairsharing.netlify.app/#/search?q=miame).
miame

What's happening with this search please? Why aren't the MIAME records being found?

thanks again!

allysonlister:
Also, as a separate point, the rotating polygons in this deploy are hovering further up than they should be - right to the top of the page, rather than in the dark blue area. Thanks!

knirirr:
> What's happening with this search please? Why aren't the MIAME records being found?

I don't know. What was in the URL field when you clicked the submit button?

allysonlister:
I thought that the duplicate check looked at record name, abbreviation or URL?

knirirr:
> I thought that the duplicate check looked at record name, abbreviation or URL?

It checks them all, if they are long enough. 
I am attempting to reproduce your search results so I can answer your question.
Please would you let me know what values were present in Name, Abbreviation and URL fields when you clicked the submit button. 

Edit: Or, were they actually empty? If so then that's a separate bug I'll have to deal with, as submission shouldn't be possible with compulsory fields empty.

allysonlister:
Sorry, I've been on a call all afternoon.

Record name and URL were empty; Record abbreviation contained "MIAME"

knirirr:
Thanks. In that case the spurious search result was a side-effect of the incorrect submission with missing fields. I've fixed that so please feel free to mark the PR as approved once you have finished looking.

allysonlister:
Other than the odd polygons in the header, this is looking great. Happy for you to proceed, thank you!!

knirirr:
As far as I can tell the particles problem is due to Github's compilation of the application.
Locally (and I expect on beta as well) the version of particles.js in package.json is respected; on Github it is not. Therefore, it may be a problem caused by building with a newer version. 
I expect therefore that it will be safe to merge.

PR #1314, opened 2021-11-02
Fixed bug with not showing icons in Record

hoseinmirian:
The problem was happening for a simple but tricky reason. Having forwardslash in front of a record ID was causing the issue. for e.g:

localhost:8080/1 was fine but localhost:8080/1/ was not fine. 
I deleted the last forward slash if available using middleware.

Issue #1312, opened 2021-11-01
Database registry logo missing from record


allysonlister:
**Describe the bug**
database logo in the top left of this record is missing, even upon refresh https://beta.fairsharing.org/3579/

**To Reproduce**
1. Go to https://beta.fairsharing.org/3579/
2. See error

**Expected behavior**
Logo to be present.

**Screenshots**
![logo](https://user-images.githubusercontent.com/143586/139714406-cad6374a-1c65-4db5-859f-7acebec08d1a.png)

**Operating system and browser version**
Firefox / Ubuntu

**Additional context**
@ramGranell confirmed this issue

PR #1316, opened 2021-11-03
Added slash to icon paths. #1312.

knirirr:
Looking at https://github.com/FAIRsharing/fairsharing.github.io/pull/1314 I thought the problem looked very mysterious.
Certainly, I had seen URLs in production of:

https://beta.fairsharing.org/123/assets/image.svg (or similar)

...which suggested to me it was a case of an image URL being supplied as `assets/img.svg` instead of `/assets/img.svg`. My modification of the icons file will achieve this.

Please have a look at this build and see if you can spot any example of the changed icons going missing. (record 3579 was the original one reported).


allysonlister:
I cannot get the problem to recur here, as hard as I try, so please do go ahead with this change. Thank you!

allysonlister (approved):

hoseinmirian (approved):
Thanks, it works fine. I close my branch

Issue #1385, opened 2021-11-24
Add additional confirmation option to curator dashboard for approvals of hidden records


allysonlister:
Whenever a record with hidden = true is approved by a curator, update the confirmation dialog from this:

![](https://user-images.githubusercontent.com/143586/143215633-2cca7d53-afb0-4ae1-913b-8c2965ce0a2d.png)

to a slightly modified version with an identical dialog, but with one additional clickable option: "CANCEL / OK / OK AND KEEP HIDDEN"

* CANCEL - cancel approval operation (same behaviour as now)
* OK - approve record and set hidden = false
* OK AND KEEP HIDDEN - approve record and retain the hidden = true (same behaviour as the OK button now)

Thanks!

knirirr:
This is awaiting review.

PR #1394, opened 2021-11-25
Rg add hidden curator dashboard#1385

ramGranell:
This is related to ticket #1385, to test go to curator panel and see the new option when approving a record in "RECORDS/EDITS AWAITING APPROVAL". Do not approve real data


knirirr (changes requested):
As far as I can tell this does what's required.
But, there's been a drop in coverage which will need fixing before merging.

knirirr (approved):

PR #1400, opened 2021-11-29
Rg add hidden curator dashboard#1385

ramGranell:
It was the same than before but know there is different behaviour when pressing accept button in the table RECORDS/EDITS AWAITING APPROVAL. The behaviour depends if the record is hidden or is not hidden. Expected behaviour is described here: https://github.com/FAIRsharing/fairsharing.github.io/issues/1385


knirirr (approved):

hoseinmirian (approved):

allysonlister (approved):
The behaviour seems to be correct, but it's hard to be sure due to the issues with "#" and links from the curator panel out to the records therefore not working properly. However, I have checked in more diagonal way and all seems well.

Please note you may wish to check that the links for the "hidden record" section of the curator panel are working as all the links there for me show "undefined" https://deploy-preview-1400--fairsharing.netlify.app/undefined. Thanks!

ramGranell:
Ok, I see that there is an issue with the link for hidden records, I will try to sort out in this ticket.

> Please note you may wish to check that the links for the "hidden record" section of the curator panel are working as all the links there for me show "undefined" https://deploy-preview-1400--fairsharing.netlify.app/undefined. Thanks!

Issue #1355, opened 2021-11-16
Change card shade for collections, standards and policies


allysonlister:
We've had some feedback that it is hard to tell that one is viewing a Collection (as opposed to other record types). While we wish to maintain consistency, we would like some relatively subtle way (in addition to the Registry/Type and icon) for a user to feel like they are viewing a collection.

I'm completely open to ideas as to how to do this, but might it be possible for the "cards" (e.g. the color in the background of General Information) to be a slightly different shade for Collections? 

I don't know if it makes sense to extend this to other record types (would it make it easier for each registry to have its own shade?), but again I would be open to what would seem to you to be the optimal solution.

Please note this is quite low priority.

Thanks!

allysonlister:
I have updated this card after a discussion on a nearly similar topic in a recent meeting. The idea is to have a slightly different (but still very light) shade for all registry types. The database registry, because it has the most records, could keep the current color, but the other three groups (collections, standards and policies) could each get a different shade that conforms to the current FAIRsharing palette.

PR #1412, opened 2021-12-08
Improve UI for registry in record

### LINKED ISSUE #1408: Find a way to tell user more clearly what registry type they are visiting
so the ticket tries to improve the UI of record display by emphasizing on the registry type of record so users would be able more quickly to know what record registry type they are visiting.

hoseinmirian:
Sticky behavior of registry type now helps the UI to emphasize to user more about the record registry type and accompanies user while scrolling until reaching the end of general information. so this way users will be more aware of the registry type and also we are not bothering users with a permanent sticky registry type. 

for testing just pick a random record and see the behavior.


allysonlister:
That is a good idea - this is in response to the discussion we had at the dev meeting about possibly changing card color, or some other solution such as this one you are showing me, to make the registry type more visible?

I won't be able to approve or otherwise until Susanna is back from her holidays and we see what she thinks, so please feel free to move onto other tasks until I update this issue next week.

Thanks!

allysonlister:
Thanks Hossein,

I've shown this to Susanna, and we agree that the sticky registry type isn't exactly what we're looking for, but thank you for the idea and for showing it to us. Please could you make something up according to #1355 and we can see how that looks? Thanks!

knirirr:
I'll close this given that this design will not be used.

PR #1417, opened 2021-12-09
Apply Allyson feedback regarding banner behavior

### LINKED ISSUE #1413: "Record claiming request has been declined" banner is visible on 404 pages, and briefly on every record page
**Describe the bug**
The record rejection/approval banner should only appear on the record relevant to that rejection or approval. However, we are noticing it in two other places on the site (@delphinedauga and I have both noticed this):
1. when viewing a 404 URL, e.g. https://beta.fairsharing.org/3333333 the banner is displayed. What is really weird about this is that I tested the rejection/approval in preview, and so I have been rejected recently - but only in preview. However, I'm seeing this behaviour in beta. Perhaps what is actually happening is that the rejection banner is shown for every 404, whether or not you have recently been rejected?
2. For every record : the red "claiming request rejected" banner appears for a second and then disappears immediately 

This is a low-priority item. 

**To Reproduce**
1. Visit https://beta.fairsharing.org/3333333
2. See the red banner
3. Visit https://beta.fairsharing.org/FAIRsharing.9kahy4 (for example, any record will do)
4. See the red banner briefly appear

**Expected behavior**
The red banner should only appear on the record the user has been rejected from

**Screenshots**

Ubuntu / Firefox

hoseinmirian:
- now must we see no banner when record is not maintained by us or either error page
- now must we see a smooth animation showing the banner to prevent the instant show of the red or green banner.


src/views/Records/Record.vue
```diff
@@ -64,24 +64,54 @@
...
+              name="fade"
+              mode="in-out"
+              appear
             >
-              <span>Your request to maintain this record has been approved.</span>
-            </v-alert>
+              <div v-if="queryTriggered">
+                <div v-if="!queryTriggered">
+                  <v-alert
+                    v-if="(!alreadyClaimed) && ((!ownershipApproved && !alreadyClaimed) && (!alreadyNotClaimed && !error))"
```
knirirr: Although this PR appears to resolve the issue, I'm finding it hard to follow this if statement. One thing that is problematic for me is that I'm not sure of the difference between alreadyClaimed and alreadyNotClaimed. I can't find a definition of this in Record.vue.
Is there anything you could add which would clarify what this does, for the sake of future maintenance?

hoseinmirian: please pull and check the comments added to clarify your concern. tnx

knirirr: Thanks for the comments. I'm afraid it's still somewhat confusing. For example, I'm not sure why there's a separate alreadyNotClaimed. Would this not be !alreadyClaimed, or is there a subtle difference I've not grasped?
The quoted code above is also rather hard to read and so it's difficult to tell what it does. Consider, for example, the long final v-if with various clauses, and the `if queryTriggered` followed immedately by an `if !queryTriggered` If you are able to refactor it to simply these statements it would help.
If not then something along the lines of:

```
<!-- At this point such and such a thing is happening -->
<div v-if="queryTriggered">
  <!-- Here, queryTriggered is checked again because... -->
     <div v-if="!queryTriggered">
     <!-- This component should only be shown if [explanation] -->
```

hoseinmirian: yes I understand, the code is complex because a banner can have 5 status and some of the banners loads before loading the page and some gets updated after query is triggered:
1- error before loading (queryTriggered=false)
2- error after loading (queryTriggered=true)
3- approve 
4- reject
5- user only claimed ownership so the orange banner will be shown.

allysonlister:
Is this ready for review again? I can't quite tell from looking at the thread... thanks!

hoseinmirian:
> quite tell from looking at the thread...

yes it is ready.

src/views/Records/Record.vue
knirirr: Indeed, it does seem rather complex. 
So, some comments in the code to let the reader know what's going on before each section would be useful. Particularly a description of `(!alreadyClaimed) && ((!ownershipApproved && !alreadyClaimed) && (!alreadyNotClaimed && !error))` (assuming it can't be simplified).

hoseinmirian: thanks sure. I ll do it.

hoseinmirian: @knirirr your request is done in my last commit. tnx

knirirr: I think that, in order to get rid of the bug in production, I'll have to approve this for now and write a separate ticket for the remaining matters. 
I've changed the name of one variable to make it a little clearer when it comes to that.

knirirr (approved):

allysonlister (approved):

PR #1418, opened 2021-12-09
New log format

### LINKED ISSUE #1415: Viewing modified logs
The changes in https://github.com/FAIRsharing/FAIRsharing-API/issues/396 will require that the record history page is updated in order to support the new format.

knirirr:
This change relies upon a complete re-working of how logs are handled, which may be found in the corresponding PR here: https://github.com/FAIRsharing/FAIRsharing-API/pull/398. I've not tagged anyone for review on that, but please have a glance at it anyway (there are considerable file changes). 

This code is running on dev-api.fairsharing.org. If it is considered acceptable then further changes will be needed to completely remove Paper Trail (see below as to why it's still there) and to modify the database accordingly. 

Please note that at present this PR does not include any code for converting Paper Trail records into the new format. This will require a separate task and may be rather difficult and inaccurate - after all, the reason behind this change is that Paper Trail was not a very effective solution for our needs. Of course, to save a significant amount of time, the logs of all edits so far during the beta period could be discarded...


hoseinmirian (approved):
Maybe would be good to also say the number in parentheses is the id of the user 
image

like: Hosein_mirian(ID:32545)

allysonlister:
Notes from checking this new code:

1. it looks fabulous! Really easy to understand!
2. I've changed every value in General Information. During this operation, which included changing the status from Ready to In Development, I got this in the log history, which says it's an "add" but which didn't get added, as I didn't deprecate the record. Do you think it was just a modification behind the scenes, e.g. from "" to null, for example?
dep
Everything else in this tab performs exactly as expected.
3. Licences and support links: Exactly as expected, thanks. Another odd deprecation reason log, this time apparently being added as "" (empty string):
reason1330
4. Publications: looks great, except the addition of the "cite this resource with this publication" flag for the lead publication doesn't seem to show in the logs:
pubs
5. Organisations & Grants: Changing an existing organisation by making it the lead organisation is not reflected in the log history. Everything else works perfectly.
6. Relations to other records: Works perfectly.
7. Additional information: Works perfectly.

Perhaps, once these small changes have been dealt with, you could ask @delphinedauga  and I both to test? Would be good to have another set of eyes before a final approval. Thanks very much!!!

allysonlister (changes requested):
See my immediately previous comment for details.

knirirr:
> See my immediately previous comment for details.

Not all of what's requested there may be possible. The first two are due to an empty field being added as the record is saved.

What are your thoughts on:

> Please note that at present this PR does not include any code for converting Paper Trail records into the new format. This will require a separate task and may be rather difficult and inaccurate - after all, the reason behind this change is that Paper Trail was not a very effective solution for our needs. Of course, to save a significant amount of time, the logs of all edits so far during the beta period could be discarded...

?

allysonlister:
> > See my immediately previous comment for details.
> 
> Not all of what's requested there may be possible. The first two are due to an empty field being added as the record is saved.

I am happy if we just see these null or "" updates, I can ignore them if that's the best solution for this.

> 
> What are your thoughts on:
> 
> > Please note that at present this PR does not include any code for converting Paper Trail records into the new format. This will require a separate task and may be rather difficult and inaccurate - after all, the reason behind this change is that Paper Trail was not a very effective solution for our needs. Of course, to save a significant amount of time, the logs of all edits so far during the beta period could be discarded...
> 
> ?

This is a trickier answer. Is there a way to "export" all existing (albeit inferior) logs in a way similar to how the legacy logs look? So we can keep whatever we captured from this beta period?

BTW, will updates e.g. by @ramGranell 's various bulk update procedures be captured by these new logs, even if the user cannot be?

knirirr:
> Is there a way to "export" all existing (albeit inferior) logs in a way similar to how the legacy logs look? So we can keep whatever we captured from this beta period?

Yes, but it will take a significant amount of time, as mentioned above. During this time I will not be available for other tickets, meetings etc. Also, the results are not guaranteed to be accurate. 

> BTW, will updates e.g. by @ramGranell 's various bulk update procedures be captured by these new logs, even if the user cannot be?

Yes.

knirirr:
> Publications: looks great, except the addition of the "cite this resource with this publication" flag for the lead publication doesn't seem to show in the logs:

There isn't such a flag. The path includes '/metadata/citations' which indicates that this is a citation, not merely a linked publication. 

> Changing an existing organisation by making it the lead organisation is not reflected in the log history.

This should now appear, but only for future edits.

delphinedauga:
Here my feedbacks concerning the log. I made some edits on this record https://deploy-preview-1418--fairsharing.netlify.app/1508 : 

- There is something strange in the history log, the order is not chronological. Edit from 6:41 is between 6:59 and 6:50. I can't reproduce this bug :/
Capture d’écran 2021-12-13 à 18 59 38

- Licence : when I add a licence, it doesn't appear on the log. Strangely it is ok for the support in the same section.
Dec-13-2021 19-30-30

- Publications: I have the same behavior as Ally: the "cite this resource with this publication" flag for the lead publication doesn't seem to show in the logs. From your answer @knirirr, it means that we should see something like this ?
op: add
path: "/metadata/citations"
value: "the_article_name"

- Organisation : when add or create an organisation, it doesn't appear on the log
Dec-13-2021 19-46-30

- Relations to other records : when create a relation, it doesn't appear on the log
Dec-13-2021 19-49-08

For the rest, it's great, and very clear, thank you !

allysonlister:
Thanks Delphine! I **really** appreciate you checking! I will just make one or two quick comments on your feedback:

> Here my feedbacks concerning the log. I made some edits on this record https://deploy-preview-1418--fairsharing.netlify.app/1508 :
> 
>  Licence : when I add a licence, it doesn't appear on the log. Strangely it is ok for the support in the same section.
> 

This isn't a bug - you're just looking in the "Edits" tab of the log history - instead, look in the "licences" tab. Remember licences are associations from a record object to a licence object... if you look in that tab, you see your edits:
licence


> 
>  Publications: I have the same behavior as Ally: the "cite this resource with this publication" flag for the lead publication doesn't seem to show in the logs. From your answer @knirirr, it means that we should see something like this ?
>       op: add
>       path: "/metadata/citations"
>       value: "the_article_name"

I _think_ I've found this in the log history:
citations

The fact that we have "added" a metadata/citations means that this publication has been added as the citation. Is that correct @knirirr ?

> 
>   Organisation : when add or create an organisation, it doesn't appear on the log

Again - as before, please look in the "organisations" tab - I think everything is there :-)

>      
> 
>    Relations to other records : when create a relation, it doesn't appear on the log

Again - as before, please look in the "associations" tab - I think everything is there :-)

Please say if you can't find anything that I think is there :-)

Thank you!

allysonlister:
All log history seems good except the point 5 from above, with respect to organisations and the "lead" flag.

I made an edit at 8:26 PM that moved the lead from SWO adminstrators to the EBI, and that added the Bodleian Library. When I looked in the organisation history, only the Bodleian Library addition was present; the removal of is_lead for SWO administrators and addition of the flag to EBI was not in the log history
islead

Everything else seems great, thanks!

knirirr:
> I made an edit at 8:26 PM that moved the lead from SWO adminstrators to the EBI, and that added the Bodleian Library. When I looked in the organisation history, only the Bodleian Library addition was present; the removal of is_lead for SWO administrators and addition of the flag to EBI was not in the log history

Indeed, the code did not support the modification of organisation_links, only their creation and deletion. I have now added the necessary changes to the data structure to support this. 

If it now behaves acceptably I should be able to deploy this to production (as discussed, abandoning recent changes in paper_trail). I will create a separate tickets for showing edits on users' pages. 

I'd also like to check that you are both satisfied that if a record is deleted then its logs will also be deleted.

allysonlister (changes requested):
> I added the organisation "100-Talent..." and removed the lead organisation from EBI, adding it to the 100-Talent... organisation. However, the log only showed the is_lead = true for 100-Talent... , but doesn't show the change from true to false for is_lead for the EBI organisation.

I'm not able to reproduce this, and tests confirm that such changes are logged correctly on the server.
Today I'll have a quick look to see if there is anything that can easily be preserved from previous changes; in any case, I aim to merge soon. 

@delphinedauga , please would you approve if you are happy.

allysonlister:
Trying again, mostly I am getting the correct behaviour for is_lead. However, as the gif for https://deploy-preview-1418--fairsharing.netlify.app/2302 shows, when I toggle is_lead for an existing organisation, the log history does not reference which organisation has been so toggled. Here's a screenshot of the log portion with the missing organisation info:
no-org-info

Here's a gif of me performing the action:
orgs

knirirr:
>  the log history does not reference which organisation has been so toggled.

Indeed - the organisation name has not been changed, and a JSON diff will correctly omit the organisation name for that reason. If you really want it there then there will be a further delay as I see if I can find a way to hack around this edge case. 
This is a different issue from whether is_lead true/false is detected.

allysonlister:
As there can be more than one lead organisation; and once set to false we won't have any way of knowing which organisation was previously set to true; I think we need an organisationid or some other identifying information for these cases.

Delphine what do you think?


--

Dr. Allyson Lister

FAIRsharing Coordinator - Content & Community


Oxford e-Research Centre, University of Oxford

Data Readiness Group: https://datareadiness.eng.ox.ac.uk/


ORCiD: 0000-0002-7702-4495
@allysonlister

Please note that I work part time. If the matter is urgent, you may wish to email ***@***.******@***.***> instead.


For more on FAIRsharing, please see https://fairsharing.org<https://fairsharing.org/>

________________________________

knirirr:
>  think we need an organisationid or some other identifying information for these cases. 

I can shoehorn something in, but it's not great.
Also, I have made an attempt at preserving what logs I can from records previously edited. These will have to be taken as they come, both in format and in any inaccuracies due to the difficulty of recreating history from old paper_trail versions. 

Please note as well that I'll have to clear out edit logs on the dev server in a moment.

delphinedauga:
> 

From my point of view, I'm not sure it is useful to have the history of the lead organisation: if a maintainer edit an organisation from "normal" to "lead", I have little way to verify this information (except to have a look on his site without necessarily finding a proof) and, by default, I will trust him, whatever the history of this information. @allysonlister do you have a user case where you find the history necessary ?

knirirr:
Meanwhile, I have used the extra time this afternoon to hack the diff to show something along the lines of what @allysonlister is after, and try to covert some old logs. Some of these are not very good but some may  contain something of use. Example:

https://deploy-preview-1418--fairsharing.netlify.app/15

If you think it worth keeping these I can do so, otherwise they can be discarded and we can stick with the new format as previously considered.

allysonlister:
Thanks @knirirr - these "hacked diffs" are very much better than nothing, so I would be happy to keep them. 

Thank you for providing organisation info for the is_lead flag. I've approved the review.

allysonlister (approved):

knirirr:
Thanks. I'll see what I can do about deploying this later today (system updates permitting) or failing that early next week.

PR #1428, opened 2021-12-14
Added link to forgotten password form. #1421.

### LINKED ISSUE #1421: Password reset link
There has been a case of a user creating an account via ORCID and then wanting to change their local password. They have found /users/password/edit, which requests the current password, and not thought to use /accounts/forgotPassword.

So, this could be solved by adding text to /users/password/edit along the lines of "Forgotten your password or haven't set it yet? Click <this>" (with a link to /accounts/forgotPassword).

knirirr:
Go to /users/password/edit to see the link.


allysonlister (approved):

hoseinmirian:
due to its an internal link routing, I replaced a tag with a router-link tag so the page will not be refreshed when navigating to the new page. tnx

hoseinmirian (approved):

PR #1430, opened 2021-12-15
Hossein fix network color #1409

### LINKED ISSUE #1409: Color node icons in relationship graph by life-cycle status
#1227 has been closed due to external library issues, and this would be a great way of introducing another important bit of information to our relationship graphs in a (hopefully straightforward) way. 

Please could the color of the nodes be changed such that they represent the life cycle status of the record? Here's the legend showing the current node color and shape:
legend

We already have a color scheme for our life cycle statuses, and this same scheme can be used. It will also require an appropriate addition to the legend letting people know what the different colors mean.

Thank you!

hoseinmirian:
Now each node color will be according to their status: (ready:green, etc)


knirirr (changes requested):
The code looks OK, but we'll need a key in the sidebar to the left to inform users which colours relate to which statuses.

allysonlister:
Thank you, this now seems to be working as expected. We now have an additional issue that wasn't clear to me until we had this build to look at.

Deprecated and Uncertain are both grey in FAIRsharing. I am talking to Susanna to see about perhaps changing one of the greys. Thanks!

knirirr:
I think this looks good (and will approve in a moment) but I presume we'll be waiting until @allysonlister hears back before it can be merged.

knirirr (approved):

allysonlister (changes requested):
Please could you add this new text underneath "Record Status", as there is no explanation of what record status is in the context of the graph.

"The colour of each point represents the life cycle status of that resource as shown in the legend below."

knirirr:
What was the result of:

>  I am talking to Susanna to see about perhaps changing one of the greys. 

Are the colours acceptable, or do we need to wait?

allysonlister:
My apologies - trying to catch up on a Friday hasn't worked as well as I had hoped :-)

Yes, she has no strong preference. However, whichever color we use for deprecated records should be the same across the site.

I would like to suggest a FAIRsharing red / maroon for deprecated, though that might be a bit bright and would have to be sufficiently different from the red color that the central node in the graph is using. Do you have any suggestions?

knirirr:
I have none, so if you don't either I'll leave this to @hoseinmirian to make changes and re-request review; I am already satisfied with everything in this PR.

hoseinmirian:
lets see how your suggestion look @allysonlister I have no preference for now.

allysonlister:
Thanks - let me know when the updated color has been implemented please :)

hoseinmirian:
actually I have an opinion. maybe brown would be good due to deprecated means somehow old and old color reminds us always like gray, black and brown mostly. old movie old images used to be gray or brownish... if that makes sense? what do u think?

allysonlister:
A shade of brown sounds good - please just make sure it is sufficiently different from "in development"'s color. Thanks very much!

hoseinmirian:
@allysonlister its ready to review. Updated the color of deprecated to brown. 

@knirirr : LightenDarkenColor function added in generalUtils due to the previous third party was causing issue with test, this one is tested and can be used in the future. Also another improvement to look is where I used the lightenDarkenColor js function to programmatically changing the gradient of record status background and also all the color now are loading from theme,js which everytime we need to change the color just need to change theme.js and it applies on record status and elsewhere.

knirirr:
@hoseinmirian looks like coverage has dropped, which will need fixing.

allysonlister (changes requested):
> please just make sure it is sufficiently different from "in development"'s color. Thanks very much!

Thanks, but the "in development" color and the "deprecated" color are now very similar. Perhaps introduce a little bit of red into the deprecated color?

tests/unit/utils/generalUtils.spec.js
```diff
@@ -0,0 +1,21 @@
+import generalUtils, {LightenDarkenColor} from "@/utils/generalUtils.js";
+
+describe("generalUtils.js", function(){
+
+    it("can check vvv", function () {
```
hoseinmirian:
```suggestion
it("can check getHostName method",function(){
```

tests/unit/utils/generalUtils.spec.js
```diff
@@ -0,0 +1,21 @@
...
+        process.env.VUE_APP_HOSTNAME = 'https://api.fairsharing.org'
+        expect(generalUtils.methods.getHostname()).toBe("https://api.fairsharing.org");
+    })
+
+    it("capitalises aaa", function () {
```
hoseinmirian:
```suggestion
    it("can check LightenDarkenColor function", function () {
```

hoseinmirian:
@knirirr and @allysonlister ready for review

allysonlister (approved):
thanks - this looks great now!

knirirr (approved):

PR #1436, opened 2021-12-20
Hossein upload logo #1332

### LINKED ISSUE #1332: Enabling logos for records
This is related to a Zulip discussion at https://pyroeis.oerc.ox.ac.uk/#narrow/stream/8-FAIRsharing-Rebuild/topic/UI.20Feedback/near/33557

Please can we begin work on enabling logos for all record types. From what was said in the Zuplip discussion, I believe the following things need to occur:

1. Creation of UI code to add/remove/update logos on a record.
2. Copying logos that are already present from the legacy (collections only) to their equivalent records on new site.
Thanks!

hoseinmirian:
UploadFiles.vue component is a fully customizable and scalable component due to the way it's been designed:

-UploadFiles.vue is responsible for taking care of the logic and completely works independent from the presentation layer.
-UploadFilesPresentation.vue is the child of the UploadFiles and only responsible for presentation, CSS, and the way upload component look. 
 - UploadFilesService.js is a file responsible to be customized and be used for uploading to different endpoints. It can get fully customized. 

1-How it works:
Add UploadFiles.vue component anywhere you like to have upload files (names in all component and classes are chosen generic) 

pass your desired props:
```
      <upload-files
        :credential-info="{id:getField('id'),token:user().credentials.token}"
        :initial-images="getField('url-for-logo')"
        :upload-service-name="'uploadOneFilePerRequest'"
        :base-api-endpoint="getAPIEndPoint()"
        :allowed-file-size-mb="3"
        mime-type="image/jpeg,image/gif,image/png"
        multiple-upload={default:false}
      />
```
Props:

- credentials: if you do have a credential then pass it as object otherwise relax and don't pass anything. it will be good when we do not need credential. component will ignore it
- Initial-Images: If you have initialImage array or string then pass it to see its preview by the first visit
- upload-service-name: this one is very important and should be a class name of UploadFilesService that handle the upload operation. (It should be a string class name)
- base-end-point: this is required and images or pdfs or any other files URLs MUST be relative. so this base-end-point prop is necessary and its required to be passed. (can be like api.fairsharing.org) and the relative image url would be '/folder/image.jpg'
- allowed-file-size: there should be a max limit of upload per each file otherwise there will be error. 
- mime-type:  you can pass any type of acceptable standard file types like pdf, txt, etc. 
- multiple-upload: if you need to select multiple files, then just include it as it is by default false. 

2- make sure your upload function is defined as async in the UploadFilesServices.js a commented example is included in this class. the name of the class is the one that will be passed as upload-service-name to uploadImage.vue component so this way we can have several upload component but each one can be connected to another class. 

Benefits of using the practice I did:
- decoupling presentation from logic by creating a parent and one child for presentation
- the upload class makes the component very decoupled and reusable as well as less opinionated as if you see I have named it 'UploadLogo' so we all know it is for uploading single image as logo. Also it is passing the data to RestClient to do the operation so its not fully tailored for our application and can be used anywhere. 
- using different props makes the component being able to be used for different types 
- presentation layout can be customized


src/components/UploadImages/UploadImages.vue
```diff
@@ -0,0 +1,197 @@
...
+     updateImageList(response) {
+          // this part cannot be more decoupled and its tailored for fairsharing application
+          // handling multiple images
+        if (isArray(response.data.attributes['url-for-logo'])) {
+          this.fileInfos = response.data.attributes['url-for-logo']
+        }
+        else if(response.data.attributes['url-for-logo']){
+          // handling single image
+          this.fileInfos = []
+          this.fileInfos.push({url: response.data.attributes['url-for-logo'].replace("api", "dev-api"), name: 'logo'});
```
knirirr: Here, the replacement an be removed. Instead, if the server simply supplies a relative URL for the image then the full URL can be constructed by means of VUE_APP_API_ENDPOINT from .env. Doing this will allow it to work properly on both beta and various dev builds, preview etc. 

e.g.

```
let url = process.env.VUE_APP_API_ENDPOINT + response.data.attributes['url-for-logo'];
this.fileInfos.push({url: url, name: 'logo'});
```

hoseinmirian:
new feature for our uploadFiles component:

- its now named uploadFiles because it's capable of uploading any file types and can be extended more... 
- if we pass multiple-upload and multiple-files-per-request prop as true to our component it will need a special service-name prop which is a function in our uploadServiceFiles that can handle this functionality. the good news is that I have created the code to handle multiple files in one request and the sample code is tested and put in the file for future use cases. 

Please note in our editor view I have put 2 uploadFiles components to show you how it works independently with multiple files per request or single file per request. After explaining and approving the branch I Will delete the extra uploadFiles component. in fact the first top one is the one that will be remained to be used. the second one is the one I created to show how it can be used with other service-name to send multiple files or any other way we want we just pass service-name method from our class and it works accordingly.

hoseinmirian:
code refactored so now everything is more and more generic. 
-uploadFilesService.js now is fully generic and even its method sendOneFilePerRequest accepts a new argument as key which is the file key that's been sent to server like "logo". it can be defined as prop name file-key-name for upload-files component to even customize it. for e.g our server accepts term of "image" instead of "logo" in other case. its easy to pass this prop and customize it. by default it is "file" as defined in the uploadFilesService.js->sendOneFilePerRequest

other props:
- title is added to show whatever we want. 
- input-lable with default value of "Select File". is added. for e.g for multiple file upload component with multiple prop I have passed input-label="Select File(s)"

other improvement:
-setFormData method in uploadFilesService.js is now more generic and accepts key and value and set anything passes to .

src/components/Editor/GeneralInformation/BaseFields.vue
```diff
@@ -431,6 +431,33 @@
...
+        :initial-images="getField('urlForLogo')"
+        :upload-service-name="'uploadOneFilePerRequest'"
+        :base-api-endpoint="getAPIEndPoint()"
+        :allowed-file-size-mb="3"
+        mime-type="image/jpeg,image/gif,image/png"
+        file-key-name="logo"
+        title="logo"
+      />
+      <v-divider />
+      <upload-files
```
knirirr: The presence of this multiple-file-upload box means that formValid isn't true unless it's filled in. Since it should never be, as only a single logo is to be uploaded, then this second upload field should be removed.

src/components/Editor/GeneralInformation/BaseFields.vue
```diff
@@ -431,6 +431,33 @@
       </v-expand-transition>
     </v-col>
 
+    <!-- Upload logo -->
+    <v-col cols="12">
+      Logo:
+      <upload-files
```
knirirr: This field causes formValid to be false unless a logo is selected. But, not every record will have a logo, so it should also be valid if empty. 
As it stands, it's not possible to update a record's GeneralInfo without adding a logo.

knirirr (changes requested):
There's a comment here with a request for displaying the logo; logos do not show on records in this branch at present and so would need to be added:

https://github.com/FAIRsharing/fairsharing.github.io/issues/1332#issuecomment-970306290

The second part (concerning re-sizing etc.) may not be feasible but It should be possible to view a logo on the record page after uploading it. 

A user may expect that editing the logo should increment changes (see the currentFields watcher in GeneralInformation.vue), with the logo not being uploaded until the save button is used. But, as configured the uploader sends the file to the record before save is clicked - not what I'd have expected. If it's possible for the file upload to be sent with the PUT containing all other data that this would be preferable, I think.

src/components/Editor/GeneralInformation/BaseFields.vue
hoseinmirian: indeed. that was added for review purposes. it l will be removed.

hoseinmirian:
> concerning re-sizing etc.) may not be feasible but It

for the first point if you agree this branch is already big and its called upload logo. so I will open another ticket to handle the front end view of the logo?

second point is related to first point then I ll do it in another branch

third part needs investigation. have to check.

src/components/Editor/GeneralInformation/BaseFields.vue
hoseinmirian: have to check it. tnx

knirirr:
> for the first point if you agree this branch is already big and its called upload logo. so I will open another ticket to handle the front end view of the logo?
> 
> second point is related to first point then I ll do it in another branch


Please go ahead create a separate branch for viewing uploaded logos, when you get chance. This one could perhaps be resolved first.

hoseinmirian:
@knirirr now uploadFiles component is equipped like a normal component with v-model and can be reactive to form. Now when there is no data selected, it will not prevent the form to be submitted. In regard to your preferred option to include it in the final save request, I tried but due to file after upload is being saved already, I think its an extra and weird reaction to ask the user to upload first and then without seeing the preview press save. I think all is set and component is ready to be used in any form with having form validation. thanks.

knirirr:
>  I think its an extra and weird reaction to ask the user to upload first and then without seeing the preview press save.

That's not what I meant. As far as I am aware the sequence of events should be something like:

1. User selects file.
2. File is loaded in the browser and displays on the page.
3. Browser sends data to the server.

At present step 3 follows step 2 immediately. I would prefer to see step 3 integrated into the form's existing mechanism for submitting data, i.e. once step 2 is complete then the logo data should be added to the data object which is then posted to the server when the save button is clicked. Does that make sense?

hoseinmirian:
file needs to be uploaded first so I can read it. when the file is on my local machine it's not readable for the browser unless we upload it somewhere and show a preview from the image URL we have received,. so the step 23 needs to be done before step 2. is not possible unless step 2 is done, besides user after save and exit without seeing the preview will be thrown out of page. so isn't better they see the preview before sending it to the server?

knirirr:
>  so isn't better they see the preview before sending it to the server?

You appear to be confirming that there is a stage at which the image is previewed in the user's browser but has not yet been posted to the API. If so, then what I am suggesting above should be possible.

hoseinmirian:
@allysonlister please share your feedback on this feature. its our new upload component added in the general information section in edit record view. Please review and write your feedback here... thanks

hoseinmirian:
new improvements:

- new props (hasPreview) letting user only preview image and then leave the saving process to save & continue button 
- same time still we have the chance to upload files directly using upload files by setting hasPreview=false so the upload becomes a direct upload and we can see the upload button.

allysonlister (changes requested):
Thank you, this is a great start! Here's what I see:
logo

Some comments about how the upload looks when you first load the edit interface:

1. I would expect the logo upload to be at the top of the General Information tab, with the name of the resource. This is because the logo itself will be displayed next to the name in the record view, so the editing of the logo should also be in a similar position in the edit interface.
2. To make it fit, up near where the name of the resource goes in the edit interface, you can make the upload section smaller, and maybe only take up 20% of the width of the page (whatever looks best). Alternatively, you could try putting the logo upload above the name / abbrev etc fields.
3. Having "Preview" as the title for the area is confusing. Perhaps change it to something like "Provide/update this resource's logo".

The functionality is great thanks. Every time I upload a new image, it really nicely updates the thumbnail and gives information about maximum size
select

Instead of the information message reading "each file size must be below 3 mb", please could you update it slightly to "images must be less than 3 mb"

Thanks very much!

delphinedauga (commented):
Agree with Ally : better at the top (near name / abbreviation)
I would also remove preview and the text below « preview images will be listed here … » and just add a title to the box. For me it’s not necessary to specify that a preview will appear, the user will see the preview. 
Logo

hoseinmirian:
@allysonlister


1- not possible due to the component being created with having a preview and it breaks the consistency of our component design. Besides, from the editor/curator's point of view, they just want to upload a logo, and the ones who see the result and most matter in terms of presentations are end-users which will be dealt with in another ticket. I only can No.2 as below and Num3 and n.4

2- I only can put the upload component above abv/name cz because of responsiveness concerns. 
3- Done!
4- correcting warning text to image files should be .. done!

@delphinedauga 
1- suggesting label of logo from you is a good idea. will be done soon. 
2- images preview text will be listed here deleted. 
3- in regard to not showing a preview, I am opposed this idea due to some technical points that needs the image to be shown before saving it. App saves all the data once its been previewed.

hoseinmirian:
new improvements:
- feedback of curators  applied as well as having a better layout in general Info in editor
- added new prop 'hasPreview' to enable us to see the preview and pass the files data using prop name passDataToParent

allysonlister:
Thank you! This is looking good. My only comment relates to how tall this logo upload box is. For that, compare the screenshot from a very wide window and a narrow one:

wide:
wide


narrow:
narrow

No matter the window size, the logo box only needs about 3/4 of the height that it is currently, as the thumbnail appears to always be a particular height no matter what you upload. Therefore could we make the logo box a bit smaller, perhaps the height of 3 of those text fields, rather than 4? Does that make sense?

Thanks!

hoseinmirian:
@allysonlister sure im working on it

hoseinmirian:
@allysonlister your comments applied as well as more improvement on the UI responsiveness.. now we have 3 column of fields in xl screens and 2columns in large and on and on.. I think it looks very good now and hope all like it! tnx

allysonlister:
Thanks! I've added a picture of me to https://deploy-preview-1436--fairsharing.netlify.app/1195 to test the new code. The image does seem to be saved, as I can see it in the edit interface. It's not yet visible in the record view or history logs yet, though. Just FYI.

I am happy with how the edit interface looks, thanks! @delphinedauga ?

hoseinmirian:
> view or history

@allysonlister @knirirr the part Ally is mentioning will be done in a separate ticket after merging this one due to this branch is already too big and its called upload photo as its been focusing on this specific task. thanks.

knirirr (changes requested):
I went through the following steps:

1. Edit record https://deploy-preview-1436--fairsharing.netlify.app/10
2. Select a new logo
3. Click "save and continue"
4. On the server:

```
2.7.3 :008 > r = FairsharingRecord.find(10)
 => #<FairsharingRecord id: 10, created_at: "2014-11-04 15:23:40.000000000 +0000", updated_at: "2022-01-19 16:11:44.926185000 +0000", metadata: {"do... 
2.7.3 :009 > r.logo.nil?
 => false 
2.7.3 :010 > r.url_for_logo
 => "/rails/active_storage/blobs/redirect/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBFdz09IiwiZXhwIjpudWxsLCJwdXIiOiJibG9iX2lkIn19--1f1dff2d34705d513e45770f4fc612270f9ae4c4/solomani.png?disposition=inline" 
```

5. Exit editing.
6. Commence editing again. 

All is good except for step 2; here, there's a regression causing the change counter on the tab to not show. 

One other (cosmetic) point; the image as, shown here, is a flag:

https://dev-api.fairsharing.org/rails/active_storage/blobs/redirect/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBFdz09IiwiZXhwIjpudWxsLCJwdXIiOiJibG9iX2lkIn19--1f1dff2d34705d513e45770f4fc612270f9ae4c4/solomani.png?disposition=inline

However, when it's previewed it's shown as a square rather than a rectangle. Ideally, the preview would show the correct shape of the flag.

knirirr (approved):
Thanks, looks good.
Looks like it will need dev merging into this branch before this PR is merged, though.

hoseinmirian:
ok thanks @knirirr then I think @allysonlister was also happy of the last changes? if yes then I ll soon merge it into dev today. tnx

hoseinmirian:
@allysonlister please review and approve the branch if you are happy so I can merge it. tnx

allysonlister (changes requested):
1. I create a record without adding a logo, all ok.
2. I create a record with adding a logo, saving the /create form doesn't actually save, just silently hangs
3. Whenever I try to add content, it hangs at saving /create. I can no longer save new records.

I'm not sure if this directly related to the logo, but as that is the main thing that's changed, I think it needs to be investigated.

Other behaviour seems to be correct, thanks

knirirr:
> 1. I create a record without adding a logo, all ok.
> 
>     2. I create a record with adding a logo, saving the /create form doesn't actually save, just silently hangs
> 
>     3. Whenever I try to add content, it hangs at saving /create. I can no longer save new records.
> 
> 
> I'm not sure if this directly related to the logo, but as that is the main thing that's changed, I think it needs to be investigated.
> 
> Other behaviour seems to be correct, thanks

For reference, I was unable to replicate (2) and (3), and had no problems creating new records.

hoseinmirian:
> * actually save, just silently hangs
> 
> * Whenever I try to add content, it hangs at saving /create. I can no longer save new

I tried it a lot with different scenarios and browsers. no problem I found.

hoseinmirian:
@allysonlister 

have you noticed the warning and after pressing I who what I am doing the create? if you press that then the create button becomes active and creating process works as usual. 
image

@knirirr I also have spotted a minor presentation issue: 
1- go to create page
2- upload a logo and enter information
3- save and will be redirected to edit record.
4- go back to create record page and still the already uploaded file was showing in the upload which is fixed in my **last commit.**

allysonlister:
I cannot save a record at https://deploy-preview-1436--fairsharing.netlify.app/create at all, with or without logo. There is no error message,it just sits there as if the "Create Record" is inactive. I can show you in person if it helps? Thanks!

knirirr:
> I cannot save a record at https://deploy-preview-1436--fairsharing.netlify.app/create at all, with or without logo. There is no error message,it just sits there as if the "Create Record" is inactive. I can show you in person if it helps? Thanks!

I am not able to replicate this and so will need to see it in person.

hoseinmirian:
@allysonlister please try it on another machine also if you have. I am also ready so you can show me in person whats happening. By next week I ll have to do the front end side done so I need this one to be resolved and merged soon. tnx

allysonlister (approved):

hoseinmirian:
@allysonlister please check this branch preview to see the record display icon changes and comment out if your happy so I can merge it to dev.tnx

hoseinmirian:
@allysonlister are you happy with this branch? please check the netlify build and let me know about it. I know that you could not see the netlify build on the other branch so that's why I merged that branch into this one so you can see the preview. I am waiting for your final feedback so I merge this branch into dev.

allysonlister:
I am happy with the updates - thanks!

allysonlister:
Is there any way to make narrow landscape-oriented logos more visible? Take this for example:

https://deploy-preview-1436--fairsharing.netlify.app/1548

hoseinmirian:
@knirirr @allysonlister 

hopefully this will be the last commit we ll have on this branch. so now I have added a condition to check if the image is rect or square but still its better to advise users before uploading in (another branch possibly) to upload images with the below sizes as I have uploaded some as examples:
square size : 100* 100 72dpi like : https://deploy-preview-1436--fairsharing.netlify.app/17
rect size : 300* 100 72dpi like: https://deploy-preview-1436--fairsharing.netlify.app/18


make sure u refresh a few time to see the changes

allysonlister (approved):
Looks great, thanks!

PR #1442, opened 2021-12-20
Fix indices in record associations

### LINKED ISSUE #1416: Removing relationships - errors
**Describe the bug**
Deleting relationships when "policies" was deselected resulted in 
1. silent deletion of policies even though they were deselected and therefore not visible, and
2. the list of relationships wasn't updated (clicking on the trash can icon resulted in the count of relationships decreasing but no apparent change to the relationship list)

**To Reproduce**
1. Go to the preview site, as this change has already happened in the beta site. Visit https://preview.fairsharing.org/3350
2. Try to delete just databases and standards by deselecting "policies" and start deleting
3. Notice the errors mentioned above.

**Expected behavior**
only remove those relationships that you choose, and update the list of relationships accordingly as you remove them.

**Screenshots**
After performing the update shown in this gif, **all** relationships to policies had disappeared, as well as a few of the standards/databases!
removing-relationships

Ubuntu / Firefox

**Additional context**
Please ping me if you have any questions, as I understand this is a complex issue. Thanks!

knirirr:
Rather than use the index of a record in the array of associated records for deletion this change compares the linked record ID and the ID of the label. There should only ever be one link for each record/label combination so this should work, and also allow deletion to occur when records are hidden. The previous bug was caused by deleting at an array index when the records' indices had changed.


hoseinmirian (approved):
seems fine

PR #1448, opened 2021-12-29
Hossein improve record UI #1408

### LINKED ISSUE #1408: Find a way to tell user more clearly what registry type they are visiting

hoseinmirian:
image

now our same old Ribbon component comes handy and I reused it to improve the Record UI and show registry type inside of it. I think its a concise and good way of emphasizing the registry type for the user. Please review and let me know if it meets your need.


allysonlister:
Thanks!

I will get back to you next week, once more people are back from their winter holidays. So, please don't do any further updates until I've collated general opinions on this idea. However, for your information, I did ask you to implement something according to #1355 (changed shades for card backgrounds) as this is something visible no matter where you are in the record. Therefore, it might well be that I ask for this again after I've spoken with others next week.

hoseinmirian:
> 

I know but I did not come with a good result with that. I think we can benefit our ribbon. please share and see what would be the feedback. then we can change it if not approved.

allysonlister:
I will share it next week when SAS is back and let you know, thanks!

allysonlister:
Thank you. Susanna and I agree that the ribbon does not provide the right kind of visual aid. Please will you work with the shading options as described in https://github.com/FAIRsharing/fairsharing.github.io/issues/1355#issuecomment-987960080

thanks!

PR #1464, opened 2022-01-12
Added link to all records for an organisation #1397.

### LINKED ISSUE #1397: add maintainer info to "records relating to this organisation" table in the organisation pages
When looking here: https://beta.fairsharing.org/organisations/1991 I felt it would be really useful to have two things:

1. A link from this table to the search page that would retrieve these same records, thus allowing the user access to the facets for this subset of records.
2. A method of knowing at a glance within the table if each record is maintained. Perhaps a dot in the same way as the record status, or even color the row a different color? If point (1) is easier, please do that and ignore this one, as one can filter on maintainer in the main search pages.

Thanks!

knirirr:
This adds a link on an organisation page to the search page for all the records in that organisation, allowing further filtering to be done.
As suggested in #1397 I've left the maintenance filtering for the search page.

### DIFF
--- a/src/views/Organisations/Organisation.vue
+++ b/src/views/Organisations/Organisation.vue
@@ -217,6 +217,16 @@
                 style="min-width: 720px"
               >
                 Records related to this organisation
+                <!-- filter records on search page -->
+                <v-col class="text-right">
+                  <v-btn
+                    color="white"
+                    small
+                    @click="filterRecords"
+                  >
+                    Filter Records
+                  </v-btn>
+                </v-col>
               </v-card-title>
 
               <v-data-table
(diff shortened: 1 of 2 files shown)


hoseinmirian (approved):

ramGranell (approved):
It seems to work great. The only comment is the button words: "Filter Records" is confusing as it seems that you are able to apply a filter over the record that are below, e.g. in https://deploy-preview-1464--fairsharing.netlify.app/organisations/1991 there are 29 records, and it seems that I can apply other filter automatically just pressing the button. Then, I will put other thing such as "Get them", "Search them", "See all of them" or something like this. Ask @allysonlister for the best wording

knirirr:
@ramGranell I thought that objection might arise, but I think my wording the most appropriate given that the purpose of the button is to enable the user to filter those records via the search facets.

Issue #1459, opened 2022-01-11
Add search in abbreviations for right-hand (existing) relations in edit interface and record view


allysonlister:
Please could we have the behaviour of the right-hand side of the "Relations to other Records" tab function in the same way as the left? Currently, typing abbreviations in the right-hand "Associated Records" menu does not find any results, while this very useful feature is available in the left-hand "Available Records".

Example: edit https://beta.fairsharing.org/FAIRsharing.q4ahrn
![ddbj](https://user-images.githubusercontent.com/143586/148956311-f59fd031-cd0b-4a75-a057-845309be7cf8.png)

DDBJ / "DNA Data Bank of Japan" is present in both menus, but only is retrieved in the left-hand one.

Thanks!

allysonlister:
One cannot search based on abbreviations in the record view either - only in the full name:
![ddbj2](https://user-images.githubusercontent.com/143586/148956535-e849fbbb-b0ec-40be-bec0-3a5268ecbec4.png)

It would be great to have abbreviations visible in these search fields on the record view as well. Thanks!

knirirr:
The search in the left column of the first screenshot queries the server, where it has the power of Elasticsearch available. The others use string matching in the browser and, though these requests can be accommodated (once I've written some tests), there's a limit on what can be done client-side (re-writing everything to use the server would be prohibitive).

allysonlister:
Thanks for the information, and it's good to know that they're not run the same way, and therefore there is a limit to what can be done here. Thanks!

PR #1465, opened 2022-01-12
Search on record page 1459

knirirr:
This implements the requests in #1459, within the constraints of what's available.


ramGranell (approved):
I can use acronyms in the right column in the editing of relationships (records that have an existing relationship with the record we are editing) and works properly.

PR #1479, opened 2022-01-21
Record delete button

### LINKED ISSUE #1477: Button for deleting/tombstoning records
Rarely, very old records need tombstoning or deleting (depending on if they have a DOI or not). Please could the Action menu for a record have a new option, only visible to the highest-level curator role (i.e. not even greyed out for other roles)?

Please duplicate the behaviour of the pop-up dialog in the curation dashboard when a curator clicks on the red "bin" icon (which will delete/tombstone as required).

Thanks!

This relates to https://github.com/FAIRsharing/FAIRsharing-API/issues/428 and https://github.com/FAIRsharing/FAIRsharing-API/issues/436

knirirr:
This adds a deletion button, accessible only to super_curators and developers, which functions in the same way as that on the curator page, to individual record pages.


hoseinmirian (approved):

ramGranell (approved):
It worked properly for me.

PR #1480, opened 2022-01-24
Distinguishing Collection type from other by changing style of genera…

### LINKED ISSUE #1355: Change card shade for collections, standards and policies

hoseinmirian:
Now in only the collection type we can see a greenish border style  tailored only for collection as well as an informative badge on the top right as well. changing background color is not reasonable to me and it breaks the look and consistency of the layout. What's done in the legacy website is not a god and is based on best practices for us! so I think the way it's now implemented is less colorful and more informative. tnx

Please check a collection record and compare it with a normal one.


allysonlister:
As discussed in the call, please will you build the cards with background shades as listed in https://github.com/FAIRsharing/fairsharing.github.io/issues/1355#issuecomment-987960080 . Thanks!

PR #1490, opened 2022-01-27
Add images for 500 and 404 instead of previous text

### LINKED ISSUE #1458: Legacy 404 (etc.) images
@SusannaSansone has requested that 404, 500 etc. messages from the old system be used in the new one as well. I've copied these images to the graphics repo: https://github.com/FAIRsharing/FAIRsharingGraphics/commit/264b5d3d4bf45b72634b7592f442c223cbda2ea8

hoseinmirian:
replaced images of 404 and 500 with the current error page we have.


knirirr (changes requested):
I can't see how the 404 looks because this goes immediately to a 500 image when I attempt to load any page. 
Also, the changes would show a 500 image when the site is set to maintenance mode. This should not happen, as a "site in maintenance" response from the server is not an internal server error.

knirirr (changes requested):
When the client connects to the server the headers in the returned data might include the information that the site is in maintenance mode. If so, then the local store will be updated with this information.
If maintenance mode is set then a mantenance message (the one removed in this latest commit) should be displayed.
But, if it tries to connect to the API and fails because the API is not reachable or returns an error, then that is the time for the internal server error image.

hoseinmirian:
maybe here is the place i got to check if the API is failing?
https://github.com/FAIRsharing/fairsharing.github.io/blob/master/src/store/introspector.js#L49

knirirr:
That's indeed the query that will get the maintenance value to store.

hoseinmirian:
I am not sure if Netlify is working as expected(it immediately goes on 500 page for some reason) but the branch is ready for review. So now what happens is:

1- if there is a maintenance value set in the header then it redirects user as already was doing to Maintannace page. 
2- if there is a miscommunication and API connection failure then it will be redirected to /errro/500 with title of server Error
3- 404 one page works as usual. 

to reproduce the above
1- set the header maintenance to true and refresh the localhost:8080
2- change the  introspecter.js line 49 as screenshot bellow to {} so we get a server error and see the 500 page:
image
3- enter a nonsense URL like localhost:8080/@ to get 404 page.

src/main.js
```diff
@@ -73,7 +73,7 @@ async function bootstrapApp() {
         await store.dispatch("messages/setMessages");
     }
     catch {
-        store.commit("introspection/setMaintenanceMode");
+        await router.replace('/error/500')
```
knirirr: This catch block is reached when `await store.dispatch("introspection/fetchParameters");` is called on line 71 of main.js. So, the result is that setting maintenance mode causes a 500 internal server error to be shown rather than a maintenance page.

knirirr (changes requested):
404 appears as expected, and 500 appears if the server is disabled.
But, there's still a 500 for maintenance mode (see comments relating to main.js).

src/main.js
hoseinmirian: the catch block did not reach when I test the code. it worked as expected. 

I intentionally set maintenance to true and then refresh the page. it does not get caught in the catch but it redirects me to maintenance page.If I cause a problem in query to get 500 internal server i get it as expected
one situation is remained and is that if maintenance is true and also there is a server error same time. so in  this case also the code I wrote caught in the catch block and redirected user to 500 due to server error now has more priority than maintenance mode. I do not see any problem. if there is please let me know or push to the branch. tnx

knirirr: I'm running your branch on my local system where I have control over the server. Turning it off causes the 500 to show as expected. Setting `maintenance = false` and all is fine. But, when `maintenance = true` the result is a 500 error. 

It's not possible to demonstrate this with the Netlify build due to the ongoing problem there.

hoseinmirian: ok thanks. let me see what I can do to fix it. tnx

knirirr: I've dealt with this problem now and will ask for a review shortly.

knirirr (approved):

ramGranell (approved):
There is a complain by Codacy but I don't know if it is relevant.

terazus (approved):

PR #1495, opened 2022-01-27
Test bug fix 1494

knirirr:
The warnings, when got rid of, revealed a problem with the test not passing data correctly (resulting in failure).


hoseinmirian (approved):
looks fine

PR #1512, opened 2022-02-04
Addresses page load hang in #1511.

### LINKED ISSUE #1511: Publication data from ORCID
Some entries are being returned without a DOI, which causes our parsing code to throw an exception when it is not found. This in turn prevents the user profile from loading. 
This was already caught on the public profile but missed on the private.

knirirr:
This applies a fix that's already on a user's public profile to their private profile, to prevent the page hanging when ORCID returns pubs with missing DOI fields.


ramGranell (approved):

src/views/Users/User.vue
```diff
@@ -458,7 +458,13 @@
...
                             obj => obj['external-id-type'] = "doi"
                         )[0];
-                        url = DOI['external-id-url'] ? DOI['external-id-url'].value : null
+                        /* istanbul ignore else */
+                        if (DOI) {
+                          url = DOI['external-id-url'] ? DOI['external-id-url'].value : null
+                        }
+                        else {
+                          url = null;
+                        }
```
terazus:
```suggestion
                        url = null;
                        if (DOI) {
                          url = DOI['external-id-url'] ? DOI['external-id-url'].value : null
                        }
```

terazus (changes requested):

Issue #1505, opened 2022-02-02
Add text search field to collection facets


allysonlister:
Currently, collection search facets have everything normal searches have except the text search field. I find myself trying to use that text search a lot lately :-) Please could we add it in?

e.g. https://fairsharing.org/3513
![searchtext](https://user-images.githubusercontent.com/143586/152133441-9a91aac3-4b61-4870-b67b-2a51cb4a9ee9.png)
Thank you!

PR #1516, opened 2022-02-08
Changed how text search box funtions so it can be used for #1505.

knirirr:
Please assure yourself that all other search boxes (e.g. at /search, homepage etc.) still function as they should, in addition to that added to a collection page (any collection will do) acting as expected.


ramGranell (approved):

allysonlister (approved):
Thank you - looks great!

PR #1523, opened 2022-02-10
Added maintainers to the edit interface. #1261.

### LINKED ISSUE #1261: Adding ability for curators to assign/remove accounts as record maintainers
Please could we have a simple interface (implemented as the developers think best) whereby curators can add and remove maintainers from records.

knirirr:
This adds maintainers to the general info. section of the editing form, only if logged in as a super curator. 
N.B. the loading of all usernames is rather slow and requires a separate ticket to resolve.


src/components/Editor/GeneralInformation/Maintainers.vue
```diff
@@ -0,0 +1,217 @@
...
+      if (!_module.maintainers.some(m => m.id === item.id)) {
+        _module.maintainers.push(newMaintainer);
+      }
+    },
+    isNew(term){
+      return !this.initialMaintainer.filter(obj => isEqual(obj, term))[0];
+    },
+    isAlreadyMaintainer(item) {
+      let _module = this;
+      if (_module.maintainers.some(m => m.id === item.id)) {
```
hoseinmirian: I think some returns either false or true so can be simplified to 
only
return _module.maintainers.some(m => m.id === item.id)

hoseinmirian (approved):

allysonlister (approved):
Fantastic, thank you so much!

PR #1533, opened 2022-02-14
Fix issue with node-sass

hoseinmirian:
I upgraded the node-sass version but not sure how to check if the problem was resolved or not. I have created this branch and checked everything in terms of the design remained unchanged. 

I downgraded the node-sass to below ver 8, so it seems it resolved the issue. 
ref:https://www.npmjs.com/package/node-sass

ref:
https://github.com/FAIRsharing/fairsharing.github.io/security/dependabot/1

explanation:
https://vuldb.com/?id.167578


knirirr:
I've fixed the problem with newer versions of tsparticles spreading all over the screen (a simple config. issue), but it now produces warnings in the test console. These are mentioned here:

https://github.com/matteobruni/tsparticles/issues/2925

knirirr (approved):

PR #1535, opened 2022-02-15
Fixing order of toggles in editor

### LINKED ISSUE #1522: Edit interface: match order of toggles for term visibility with the term blocks just above
One of our community curators has highlighted the fact that the order of the toggles in the setting section for our terminologies (in General Information tab) is different from the order of the term blocks just above it (see image below)

image001

Please could we keep the current order of the term blocks (top part of image) and update the order of the toggles to match?

Thanks!

hoseinmirian:
fixed order of toggles in editor to be matched with the keywords in general Information section.


hoseinmirian:
for @allysonlister to check by screenshot!:
image

knirirr (approved):

allysonlister (approved):

PR #1540, opened 2022-02-16
Changed approval controls for hiding and reviewing records. #1483

### LINKED ISSUE #1483: Updates to Record Reviews
* A new, additional "OK & Reviewed" option when the green tick is clicked in the approval queue on the dashboard, making the full option list Cancel, OK, OK & keep hidden, OK & Reviewed. [The curator dashboard's approval button should no longer transmit create_review when approving a record unless the last option is selected].
   * All records should display a last curation review date to curators in whatever way the devs would like.

This relates to https://github.com/FAIRsharing/FAIRsharing-API/issues/423

knirirr:
This replaces "OK and keep hidden" and a previously suggested "OK and create review" with a toggle for hiding records (red/on = hidden) and a checkbox to create a review. These must be where they are currently located.


ramGranell (approved):

allysonlister:
This is working as expected, thank you. 

However, I find myself unreasonably wishing that either both are tickboxes or both are toggles. I have no preference as to which is done, but somehow find one of each oddly uncomfortable. Do you mind making them both the same, whichever one you prefer?

Thanks!

knirirr:
I prefer them as they are (obviously), but I'm not the one who's going to be using them. I think that a pair of switches would be the least unappealing and so will try that when I have time to get back to this.

allysonlister:
Thank you :)

PR #1548, opened 2022-02-19
Refactor and extended RestClient Class to CRUD class

hoseinmirian:


hoseinmirian:
This branch is created to improve and shorten the code of RESTclinet to make it more versetile and reusable. 
I have created a generalized class name "CRUD" which can take care of all CRUD operation in any app. this class also has method to set auth and can have more. this branch has reduced the code lines of RestClient and also has nothing to do with the functionality of the app but just refactor the code to be more organized and scalable.  

RestClient class is now extended from  CRUD class to keep its code shorter and more legible..

the CRUD class has four method:
Create for posting data
Update for updating data
Delete for deleting data
GET for fetching data

CRUD class methods names are chosen generic as it is doing CRUD operation and even can be used individually in any other asyncronous app since its not abstract then it can be instantiated even individually.

knirirr:
This can be set aside for future consideration.

PR #1550, opened 2022-02-21
Add back-button

### LINKED ISSUE #1519: Please add a "back to record" link/button from the graph pages
From a user (thanks!):

Please could we have a “back to record” link or button from a relation graph? This would be helpful when navigating from one record to another through the graph.

This "back to record" link would go to the record that corresponds to the graph, and not e.g. to the record that the user started navigating from if they have since clicked through several graph nodes.

Of course, you can work around it by deleting  «/graph» from the URL (e.g., https://fairsharing.org/graph/2110 > https://fairsharing.org/2110). But there’s no connection between that syntax and the record’s DOI or site URL (in the example above: https://doi.org/10.25504/FAIRsharing.cnwx8c and https://fairsharing.org/FAIRsharing.cnwx8c).

hoseinmirian:
image


knirirr (changes requested):
It works, though I'm not keen on the location as it produces these effects at the top of the graph element:

Screenshot 2022-02-21 at 14 22 33
Screenshot 2022-02-21 at 14 22 11

Could it be fitted under the text instead?

hoseinmirian:
no its not possible to place it below text. I tried to give it a abseloute position but its not working besides there are some cases with a lot of graph nodes that makes the page very packed... It was the best place to put it.

knirirr:
>  It was the best place to put it.

What can do you about the cut-off top edge of the graph with it in that location?

hoseinmirian:
@knirirr how about this?

image

knirirr:
I prefer the effect if the following is done:

```
 chart: {
                plotBorderWidth: 0,
                plotShadow: false,
                ...
}
```

But, I suggest letting @allysonlister take a look next week.
Screenshot 2022-02-22 at 12 31 46

knirirr (approved):

PR #1552, opened 2022-02-23
Assigning user roles 1542

knirirr:
This is supposed to add the "role" field to user profile editing so that curators may change users' roles.
However, it can't be merged as I can't get the tests to run. In particular, the RESTclient's `getUserRoles` function never appears to be covered.


knirirr:
As I'm not likely to ever resolve the problem here I've worked around it.
Please @allysonlister and @delphinedauga , have a go to see if you can change users' roles when logged in to your admin accounts but not if logged in as an unprivileged user.

delphinedauga (commented):
Good morning Milo !
When logged as delphine.dauga, I can edit a profile and access to the "role" options, but there is currently no defined role ("no data available"). When logged as a beta user, I can't edit a profile. Does it answer your question ? 
Capture d’écran 2022-03-23 à 09 29 13

knirirr:
@delphinedauga I have found what broke since the last time I dealt with this, and fixed it.

ramGranell (commented):
It seems that I'm now able to modify the role of the user but I didn't want to save the data as I have two questions: 1) if I modify its role or other user field, is the user going to receive any email with these modifications? and 2) if there is any kind of notification, is this happening only in production or also when using the dev DB that uses Netlify? Thanks

knirirr:
There are no notifications and this will affect only the development database.

ramGranell (approved):
It works as expected.

delphinedauga (approved):
Ok ! My account can edit the role of users ; a user account cannot edit another user profile.

allysonlister:
I confirm that the role pulldown is unavailable to my plain user (allyson_user), but is available to my admin role.

Please note that the URL for the user's page doesn't work in the netlify builds; not sure if that's something worth fixing or not.
undefined

allysonlister (approved):

knirirr:
> Please note that the URL for the user's page doesn't work in the netlify builds; not sure if that's something worth fixing or not.

That's a known problem; last time I looked at it I couldn't find a means to fix it.

knirirr:
This will make its way to the production server next time I merge dev -> master, which will probably be once https://github.com/FAIRsharing/fairsharing.github.io/pull/1607 is dealt with.

PR #1553, opened 2022-02-23
Hossein refactor banner #1425

### LINKED ISSUE #1425: Record.vue refactoring
As a result of the iterative addition of banners to the record page the logic for displaying them has become rather convoluted and hard to understand/maintain. Therefore, this should be refactored at the earliest opportunity.

hoseinmirian:


hoseinmirian:
I have categorized banner into 2 cats: 1- blue alert which are not needing any authenticated user and its global and orange banner to only show to cuartors. Besides there is one special one which is only red or green and it is for ownership approval.

so the way alert builder works is to stack all the banners and show them based on the order of being called.

**
I need help curators to identify the below conditions for the banners:
1- please tell me which ones are global?
2- please identify which one needs user being logged in. 

        .isAwaitingApproval()
        .isWatching(this.isWatching())
        .isNeedingReview(this.needsReviewing())
        .isNeedingReviewAndBeenReviewed(this.reviewsPresent())
        .isAlreadyClaimed(this.alreadyClaimed)
        .isHidden()
        .isOwnerShipApproved(this.ownershipApprovalStatus)


AlertBuilder class is based on builder/chain design pattern and can be extended more in the feature. Order for collecting banner only matters when creating the class but does not matter in the class where to define them but I have also commented out to guide where to put each type of function.

hoseinmirian:
new improvements:
- snackbars simplified and easily can be scaled. 
- Record context menu pushed into a separate component but the logic remained in Record
- Banner now is tested and ready to be used. and Record got very much simpler than before and refactored

knirirr (changes requested):
I had a quick look at this and didn't see banners for:

- Having already requested a record.
- Watching a record.
- Needs review from curators (curators only banner).
- Record isn't approved.
- Record is hidden. 

This is unexpected as a look at the code suggests that it should be OK. 
Does anyone else see the same?

hoseinmirian:
@knirirr sorry I think one stupid code movement happened in my last commits. Let me fix it and I ll be back 

All the point u made should be working! that's why branch was made.

hoseinmirian:
should be working now. tnx

knirirr (approved):
Thanks - looks good, banners appear to be working as expected now.

allysonlister:
Just looking now. Do you still need us to categorise the banners for you? If so, please provide a list of all current banners and I can do this. Thanks!

hoseinmirian:
@allysonlister no its not needed. its done. please review and approved the branch if happy!

allysonlister:
no-banner
Hello! Just edited the Subject Ontology record (https://deploy-preview-1553--fairsharing.netlify.app/1405) and no banner has appeared stating that it is waiting for curator approval.

Similarly, I've tried claiming  https://deploy-preview-1553--fairsharing.netlify.app/FAIRsharing.dg76vs and there's no banner saying that I've submitted my claim. 

Am I missing something? Thanks!

hoseinmirian:
@allysonlister Is there any concern? I think Milo answered u in zulip. So if you were happy with the branch please approve so I merge it. tnx

allysonlister:
Yes, the banners don't seem to be appearing when I expect them to, as I mentioned above.

hoseinmirian:
@allysonlister  I. did check on netlify all good!its strange!

image

allysonlister:
OK, so I can see a few things here.

1. "ownership request" message: Your account (Hossein) is showing that you are the one who requested ownership. It is actually my account that requested ownership - I should be seeing the message, not you.
2. I'm a curator, so I should be seeing the "periodic review" message, but I'm not
3. "record awaiting review" should be visible to everyone, even if not logged in, and it's not visible to me.

hoseinmirian:
@allysonlister would you please explain all the expected behavior with auth and without so I can resolve the issue. tnx

1- its another record I personally requested, please note the record Name

allysonlister:
Right, if you need all of the expected behaviour for all of the banners, and couldn't get it from the existing code, then you will need to supply me with a full list for:
* all banner texts
* all of the different banner possibilities (color, permissions etc)
Otherwise I cannot provide you with what you're asking.

You said you had this information already? It should be present in the existing code that you were refactoring.

knirirr:
> You said you had this information already? It should be present in the existing code that you were refactoring.

In a manner of speaking; the template for the page was such a mess that it was very hard to determine what was supposed to happen by looking at the code.  Plus, as far as I am aware, there's no written specification.

Concerning this PR, I'm running the same branch locally. Compare these two screenshots of me viewing a record where I've set approved to false, both logged in and out:


Screenshot 2022-03-02 at 08 49 11
Screenshot 2022-03-02 at 08 48 33

This is how I took a look at the results; all seems to be in order. Apparently, for the Netlify build, that isn't the case. @hoseinmirian , I presume you're developing locally but connecting to a remote development server, whereas I have both a local application and local server. If we're both seeing expected behaviour then perhaps there's a problem with the Netlify build - not that I can think what it might be at present.

hoseinmirian:
@knirirr  thanks for the clarification. So what do you suggest to @allysonlister to do? This is my .env file content:
VUE_APP_API_ENDPOINT="https://dev-api.fairsharing.org"
VUE_APP_HOSTNAME="https://beta.fairsharing.org/"
VUE_APP_MODE="history"

knirirr:
I added a debugging line to `checkAlerts()` (lines 416 and 417). When viewing a record with `isApproved: false` I see the following locally:

`ALERTS: {"isAwaitingApproval":{"type":"info","message":"This record is awaiting review by FAIRsharing curators"}}`

...but on Netlify:

`ALERTS: {}`

So, `AlertBuilder` isn't building alerts on Netlify. 
Currently I don't know why there's such a discrepancy, but it might be an avenue to pursue.

hoseinmirian:
no way to ignore netlify? If its working correctly then its fine

knirirr:
What will happen when it is deployed into production? At present we don't know.
Also, some features simply not doing what they should here will make it rather hard to develop them further. So, sooner or later, an answer will be needed.
I would suggest looking further into the AlertBuilder in a similar manner to what I've done above.

hoseinmirian:
@knirirr 

my netlify prints and show as expected: please check the console and result in the web page.

image

knirirr:
> my netlify prints and show as expected: please check the console and result in the web page.

Still no banners and an empty alerts object for me on Netlify, I'm afraid.

allysonlister:
Looks like I'm still waiting for now? Please let me know when I should check.

Thanks :)

hoseinmirian:
> Looks like I'm still waiting for now? Please let me know when I should check.
> 
> Thanks :)

I don't see the problem in my netlify the more I tried the less I saw any problem. not sure how should I test it.

knirirr:
I have finally tracked down the cause of why this behaved erratically on Netlify.
@allysonlister , @delphinedauga if you wish to take a look please do so.

hoseinmirian:
@knirirr tnx for fixing it. what was the cause of the issue? which line of code if u point me to. tnx

knirirr:
> @knirirr tnx for fixing it. what was the cause of the issue? which line of code if u point me to. tnx

A bug related to the calculation of record review dates (`isBannerExpired`) which was hidden by a try/catch block with an empty catch.

allysonlister (approved):
Looks great - thank you very much!

delphinedauga (approved):
Nothing to add, sounds ok !

PR #1565, opened 2022-03-01
Hossein hardcoded additional fields #1253

### LINKED ISSUE #1253: Hardcoded additional fields
The additional info section on the record page was meant to be handled in a similar way to the equivalent area of the edit form, e.g. a query to the server to see what fields were available, and then the construction of view elements based on the properties of those fields.

https://github.com/FAIRsharing/fairsharing.github.io/blob/master/src/components/Editor/AdditionalInformation/EditAdditionalInfo.vue

Instead, it's got a lot of hardcoded field names which are rather difficult to edit when the schema specification is changed. 

https://github.com/FAIRsharing/fairsharing.github.io/blob/master/src/components/Records/Record/AdditionalInfo.vue

Not such an issue if the schema specifications are relatively static (as was thought) and time was short. When time is more plentiful, it would be worthwhile refactoring this as originally intended.

hoseinmirian:
Refactored AdditonalInfo component so now it's easy to maintain and scale. 

- AdditionalInfo.vue completely refactored and code is shorten.
- new component name "DataSetArray.vue" added to take care of each category of data....


knirirr (approved):
This looks much better, although I'd like @allysonlister or @delphinedauga to take a look when they have a moment to check they are happy with the display of the additional info. section of records.

Example:

https://deploy-preview-1565--fairsharing.netlify.app/FAIRsharing.hq12h8 (this branch)
https://fairsharing.org/FAIRsharing.hq12h8

delphinedauga (changes requested):
The title should be "COS TOP Guidelines" instead of "Cos Top Guidelines"

hoseinmirian:
> The title should be "COS TOP Guidelines" instead of "Cos Top Guidelines"

I dynamically read the titles coming from API. and clean and capitalize them. Not possible to customize it cz I keep reading them from API.

knirirr:
> > The title should be "COS TOP Guidelines" instead of "Cos Top Guidelines"
> 
> I dynamically read the titles coming from API. and clean and capitalize them. Not possible to customize it cz I keep reading them from API.

`cleanString` in stringUtils.js could incorporate a hack to detect and modify this specific string by means of a lookup table.
This is the only appropriate means; unfortunately it is not particularly elegant.

hoseinmirian:
the problem is not just this tite. we have loads of title. if I be given a set of rules like make the first 2 words CAPITALIZE I can but we have different titles with different lengths and number of words. like deprecation reason, etc....

so please define some rules based on the explanation so I can follow. tnx

hoseinmirian:
@delphinedauga please check now.tnx

knirirr:
I've lost the ability to comment directly on the code; no idea why.
Anyway, this is the sort of thing I had in mind. I'd change `setCOS_title` to `setTitle` though; any further special cases could be added to this later, if necessary adding a JSON data structure to handle them.
Alternatively, changes to the schemas (if a separate title field can be supported) might be an option.
In any case, that's work for a later date as the code in this PR is (I think) significantly more maintainable.

src/components/Records/Record/AdditionalInfo/DatasetArray.vue
```diff
@@ -0,0 +1,179 @@
+<template>
+  <div
+    v-if="currentField && currentField.length > 0"
+    class="pa-4 mt-4 data-holder"
+  >
+    <b class="text-h6 text-capitalize">{{ setCOS_title(cleanString(title)) }}</b>
```
knirirr:
```suggestion
    <b class="text-h6 text-capitalize">{{ setTitle(cleanString(title)) }}</b>
```

src/components/Records/Record/AdditionalInfo/DatasetArray.vue
```diff
@@ -0,0 +1,179 @@
...
+    currentField: {default: () => [], type: Array},
+    currentKey: {default: null, type: String}
+  },
+  computed: {
+    getCurrentKey() {
+      return this.currentKey;
+    }
+  },
+  methods: {
+    setCOS_title(title) {
```
knirirr:
```suggestion
    setTitle(title) {
```

knirirr (approved):

src/components/Records/Record/AdditionalInfo/DatasetArray.vue
knirirr: Looks like the code editing button has reappeared.

tests/unit/components/Records/Record/AdditionalInfo/DatasetArray.spec.js
```diff
@@ -0,0 +1,36 @@
...
+        wrapper.setProps({currentKey: "data_curation"})
+        expect(wrapper.vm.getUpdatedTypeTitle()).toMatch("Steps");
+        wrapper.setProps({currentKey: "data_deposition_condition"})
+        expect(wrapper.vm.getUpdatedTypeTitle()).toMatch("Restrictions");
+        wrapper.setProps({})
+        expect(wrapper.vm.getUpdatedNameTitle()).toMatch("Name");
+        wrapper.setProps({currentKey: "resource_sustainability"})
+        expect(wrapper.vm.getUpdatedNameTitle()).toMatch("Plan");
+        let title = "cos top guidelines"
+        expect(wrapper.vm.setCOS_title(title)).toMatch("COS TOP Guidelines");
```
knirirr:
```suggestion
        expect(wrapper.vm.setTitle(title)).toMatch("COS TOP Guidelines");
```

tests/unit/components/Records/Record/AdditionalInfo/DatasetArray.spec.js
```diff
@@ -0,0 +1,36 @@
...
+        title = "anything else"
+        expect(wrapper.vm.setCOS_title(title)).toMatch("anything else");
```
knirirr:
```suggestion
        expect(wrapper.vm.setTitle(title)).toMatch("anything else");
```

delphinedauga (approved):
Perfect ! Thanks !

allysonlister (approved):
I'm not sure how completely this changes fields in Additional Information, or if it's just for the COS TOP Guidelines, but I'm happy with the way it looks now based on the example that @knirirr asked us to check only. Please note that in future, COS might get back to us about how they would like this section displayed, so future modifications might be requested.

PR #1566, opened 2022-03-02
Apply Ally feedback

### LINKED ISSUE #1564: Updates to back button in graphs
Relating to #1519 

As requested in #1519, there are two parts of that ticket that have not been implemented:

1. The button should read "Go to Record" (this is a slight update on my original "Back to Record" to make things clearer)
2. The behaviour should be:
> This "back to record" link would go to the record that corresponds to the graph, and not e.g. to the record that the user started navigating from if they have since clicked through several graph nodes.

This will therefore duplicate the functionality of the red centre of the graph, as @knirirr mentioned.

Thank you!

hoseinmirian:


allysonlister (approved):
Thank you, looks and behaves great :)

knirirr (approved):

PR #1569, opened 2022-03-03
Apply Delph feedback

hoseinmirian:


delphinedauga (approved):
Ok !

knirirr (changes requested):
Unfortunately, though this fixes the immediate problem, legacy.fairsharing.org will eventually disappear and so this image will become unavailable.
So, it will need to be moved into the new application. 
Good that it's been spotted before we turned legacy off, though.

hoseinmirian:
ok I download the file and push it in our public/assets ? @knirirr

knirirr:
That's probably the best place for it, yes.

knirirr (approved):

PR #1583, opened 2022-03-09
Fiddling with relations

### LINKED ISSUE #1579: Relationships to Collections should only be displayed IN COLLECTIONS section
**Describe the bug**
 I just noted that https://fairsharing.org/FAIRsharing.w4bthd has the Chemistry collection in both "OTHER RELATED RECORD(S)" and "IN COLLECTIONS". 

**To Reproduce**
1. Go to https://fairsharing.org/FAIRsharing.w4bthd
2. See presence of "chemistry" in both sections (as screenshot). 

**Expected behavior**
All collects relationships should only be present in IN COLLECTIONS

**Screenshots**
collections-and-others

Firefox / Ubuntu

### LINKED ISSUE #1580: relationships to policies should only be displayed in the IN POLICIES section
**Describe the bug**
Policies that are "in" other policies are currently displayed within OTHER RELATED RECORDS, when they should be shown in IN POLICIES.

**To Reproduce**
1. Go to https://fairsharing.org/FAIRsharing.w4bthd
2. See screenshot for example of policies found in related records section.

**Expected behavior**

Let's take as examples the following:
1. PeptideAtlas https://fairsharing.org/1927
2. Springer Nature - Research Data Policies https://fairsharing.org/3350
3. Springer Nature Research Data Policy Type 3 - https://fairsharing.org/3426

IN POLICIES should contain:
* **only for database and standards records**: any "recommends" relationships that record is a part of (e.g. "Springer Nature - Research Data Policies" recommends PeptideAtlas") - this is already happening, and is correct. I've included it here for completeness.
* **only for policy records**: any "extends" relationships that policy is a part of, e.g. _Springer Nature Research Data Policy Type 3_ extends _Springer Nature - Research Data Policies_ should be visible within the  _Springer Nature - Research Data Policies_ IN POLICIES section.

Please note that this ticket does not relate to the "related to" relationship that exists between all record types including policies (you can see "related to" between two policies if you look at this example https://fairsharing.org/FAIRsharing.tsp58); no changes need to be made to this relationship type.

**Screenshots**
collections-and-others

Firefox / Ubuntu

knirirr:
This deals with both the bug in https://github.com/FAIRsharing/fairsharing.github.io/issues/1579 and the change request in #1580. 

@allysonlister and @delphinedauga please check a few records to see that this now matches your expectations.

@hoseinmirian please take a look at the code (this modifies some of your earlier work).

Thanks.

Edit: I'll assign reviewers once the slow build/scan finishes.


allysonlister:
Thanks. The collections work looks good. I have a question about the policy work though.

Compare https://deploy-preview-1583--fairsharing.netlify.app/FAIRsharing.w4bthd with https://fairsharing.org/3416 please.

There are 14 data policy relationships in https://fairsharing.org/3416. According to the new rules, 13 of them (all except "springer nature - research data policies") should be in the IN POLICIES, and then "springer nature - research data policies" should be in OTHER RELATED RECORDS. Instead, for some reason only 1 policy relationship is visible here.

If this rule is too difficult, I propose a modification to it:
1. Change the label of IN POLICIES to POLICY RELATIONS
2. put any type of policy relationship in that section (_recommends_, _extends_ when a policy is at the other end of the relationship, _related_to_ when a policy is at the other end of the relationship)
3. OTHER RELATED RECORDS becomes perhaps unnecessary - what relationship types would be left in there?

This alternative might be altogether more simple. Thoughts? @delphinedauga and @knirirr ?

knirirr:
> Instead, for some reason only 1 policy relationship is visible here.

Sorry, I don't know what that reason is. Here's what I see on my development system:

Screenshot 2022-03-09 at 12 53 14

> This alternative might be altogether more simple.

I think this would be a bit complex. I am already confused by this part of the code and the display it produces.

allysonlister:
could there be a difference in the data on the version I'm looking at?

allysonlister:
> I think this would be a bit complex. I am already confused by this part of the code and the display it produces.

The reason I propose it is that, a user might wonder why some relations to policies are in IN POLICIES and others are in OTHER RELATED RECORDS. My alternative solution would solve this, putting anything where the "other end" of a relationship is a policy into one section.

knirirr:
> could there be a difference in the data on the version I'm looking at?

It's possible, but I don't know yet. 

I'd really rather not make any further changes to this part of the system.

allysonlister:
At a minimum, I can't approve this until we can work out why the example I've tested isn't showing what it should. Can we refresh the data that this build points to?

While I do apologise for not suggesting this alternative from the start (it wasn't clear that it would look the way it does until I saw the changes you made), I would prefer to proceed with it, and have all policy relationships in one place. However, if you don't want to do this now, I understand. I can't guarantee it won't become a ticket in the future, however, should users notice the odd disparity.

knirirr:
> Can we refresh the data that this build points to?

I can, and I am attempting to do so. 

>  I would prefer to proceed with it, and have all policy relationships in one place. 

Please write a separate ticket for that; I'd like this one to be reviewed on how it met the original request.

hoseinmirian (approved):
seems ok to me.tnx

knirirr:
I have completed a restore of the latest data to the dev server, but we'll have to wait for Netlify to come back online in order to carry on. 

Edit: @allysonlister @delphinedauga Netlify is back.

allysonlister:
This fixes the bug in collection relationships displaying and correctly updates how policy relationships are displayed according to #1580. However, due to ongoing discussions in #1584, I suggest not merging this to master until #1584 is resolved, as it will likely change how policy relationships are shown yet again (my apologies, but the question of policy relationship display has become a tricky one).

knirirr:
>, I suggest not merging this to master until #1584 

#1579 needs to be fixed, so I will abandon this PR and create another which only deals with that; further changes can then be dealt with separately.

Issue #1568, opened 2022-03-03
Adding a url field to the "mailing list" category support


delphinedauga:
Currently, the fields describing the "mailing list" category (support section) looks like this:
<img width="692" alt="Capture d’écran 2022-03-03 à 11 47 40" src="https://user-images.githubusercontent.com/59688426/156549770-1de2b7fe-8e19-4c44-9237-6c0e9e17bb43.png">

It's nice to have the email field : the mailing list email (e.g. lido-wg@lists.uni-marburg.de) allows a user to send an email and all the subscribers of the list will receive the email.
But the mailing list email doesn't allow a user to subscribe to the list. For that, a link is necessary (e.g. https://www.lists.uni-marburg.de/lists/sympa/subscribe/lido-wg?previous_action=info).

The two informations are nice to catch (email + link). Thus, we propose to add a new field for the url (optional): 
name (text) + email address (email format) + mailing list information page (url format)

- That wouldn't require any re-annotating of existing records, but of course will require updates to the record edit and record views.
- It will also require changes to the relevant schema file to add the additional field. The added field will have to be optional to avoid invalidating existing records.

knirirr:
The schema doesn't support this, I'm afraid. Changing it to do so in a useful way would require modifying the metadata of all records, and some changes to the client.

But, I note that the schema permits either a URL or an email to be supplied for a support link, but the client side validation permits it to be only one or the other depending on the field (I haven't yet worked out how this is coded).

If I could change it to allow either a URL or an email address in the form, would that be acceptable, at least until there's time for a more complex redesign?

delphinedauga:
Allowing the 2 possibilities (mail OR url) is OK for me. 
For curation, I will certainly use the url instead of the mail, as when a user will go to the url, he will access both info (the email to send to the mailing and the firm to subscribe to the list).
@allysonlister do you agree?

allysonlister:
I agree @delphinedauga - let's progress as @knirirr suggests. Thanks!

PR #1601, opened 2022-03-17
new rules and code for case when valid email or url for support link

ramGranell:
This is related to ticket #1568 . To test it edit any record and create a support link of type "Mailing list" that now accept email or url (not empty). In the code I needed to create a new rule to obtain the correct label when the pattern is not an email or a url.


knirirr (approved):

allysonlister (approved):
this works as expected, thanks!

PR #1602, opened 2022-03-17
Added link to API in page header.

knirirr:
This adds a link in the html `<head>` with a link to the API, e.g.

`<link rel="describedby" type="application/json" href="https://api.fairsharing.org/fairsharing_records/2">`
`<link rel="cite-as" type="application/html" href="https://doi.org/10.25504/FAIRsharing.2mk2zb">`

The example above is for the record that would be reached via https://doi.org/10.25504/FAIRsharing.2mk2zb


knirirr:
Codacy is applying rules differently here and locally; specifically, the "correct" location of the `head:` block relative to e.g. `computed:`.

ramGranell (commented):
In the case the DOI exists, it seems correct. However, does it the expected content when there is not DOI as in https://deploy-preview-1602--fairsharing.netlify.app/3344 ?:
"links rel="cite-as" type="application/html" href="NaN""

knirirr:
> In the case the DOI exists, it seems correct. However, does it the expected content when there is not DOI as in https://deploy-preview-1602--fairsharing.netlify.app/3344 :

The expected content then would be a citation link to the FAIRSharing page itself, as we do on the record page. What do you see? If it's something like NaN for the link then that's a Netlify problem; run the branch locally and a proper link should be generated.

ramGranell:
> > In the case the DOI exists, it seems correct. However, does it the expected content when there is not DOI as in https://deploy-preview-1602--fairsharing.netlify.app/3344 :
> 
> The expected content then would be a citation link to the FAIRSharing page itself, as we do on the record page. What do you see? If it's something like NaN for the link then that's a Netlify problem; run the branch locally and a proper link should be generated.
In both netify and locally I see the same for record 3344:

"\<links rel="cite-as" type="application/html" href="NaN">\</links>\"

ramGranell (approved):

Issue #1605, opened 2022-03-23
Colour the ADD CONTENT and STATS buttons in the header differently


allysonlister:
As has been done with the LOGIN button, please can ADD CONTENT and STATS be coloured differently than STANDARDS, DATABASES POLICIES and COLLECTIONS? This is to make it clear that they have different functions than the "quick search" buttons.

I realise that FAIRsharing blue is already used for which page you have clicked on (e.g. when you click on ADD CONTENT and are viewing https://fairsharing.org/new that the ADD CONTENT button is blue), and I don't think you want to use the FAIRsharing maroon as that's in use for our LOGIN button. Further, I also think none of us want a rainbow in the header, so I'm not exactly sure how to resolve this, and am open to suggestions as to the actual colour used.

Thanks!

ramGranell:
As a suggestion and easy implementation (all buttons are created in a [loop](https://github.com/FAIRsharing/fairsharing.github.io/blob/master/src/components/Navigation/Header.vue#L33) and therefore they have the same behaviour: background colour when not clicked and selected colour when clicked), I suggest to modify only the border colour when not clicked as this image:
![suggestion_colours](https://user-images.githubusercontent.com/39826987/159941208-8636cd4e-10da-446e-8cb8-4d32ba953d14.png)
Later when clicked the colours will be as these two images:
![suggestion_colours_add](https://user-images.githubusercontent.com/39826987/159941389-da71e96e-f6d1-47f4-8a8c-261d98f34ab1.png)
![suggestion_colours_stats](https://user-images.githubusercontent.com/39826987/159941408-f9c74e89-f391-4371-9961-2eb416029dbb.png)
The rest of buttons have same colours as before.

allysonlister:
Thank you! I'm not sure if the colours are right yet. Please could you change the orange and purple to light grey in the middle and a dark grey outline? Thanks!

allysonlister:
It might help to look at the colour for the ADD CONTENT and STATS in the legacy site...
![image_2022_03_24T17_48_30_396Z-1.png](https://user-images.githubusercontent.com/143586/159981352-27499aa0-30b0-41ec-8435-7c7035c0c383.png)

ramGranell:
Purple colour is great for statistics!!!! Meanwhile grey is a little depressing... anyway,
[...]

PR #1613, opened 2022-03-25
colour modified in the header

ramGranell:
This is based on #1605. and the colour of two of the buttons are modified: grey border when not clocked and completely grey when clicked.


knirirr (approved):

allysonlister (approved):
looks great, thanks!

Issue #1615, opened 2022-03-29
Failure to add organisations when editing records


knirirr:
A problem is discussed here:

https://pyroeis.oerc.ox.ac.uk/#narrow/stream/13-FAIRsharing-Development/topic/Investigating.20site.20issues/near/38458

It appears to be that organisations are not immediately made available for attaching to records after their creation (still requires investigation though).

PR #1616, opened 2022-03-29
Dealt with #1615.

knirirr:
This was a side effect of; https://github.com/FAIRsharing/FAIRsharing-API/issues/476


ramGranell (approved):
I was able to create a new organisation and assign it to the the record. All seems to work properly.

allysonlister (approved):
Works as expected, thank you!

Issue #1618, opened 2022-03-30
Tools section remains even when a tool is removed.


allysonlister:
**Describe the bug**
I removed the only associated tool in a record (https://fairsharing.org/3968) and the empty box for Tools remained in the record view.

**To Reproduce**

1. Go to https://fairsharing.org/3968
2. Scroll down to Associated Tools
3. Empty tools box remains. 

**Expected behavior**
The card for associated tools should disappear if all of the tools are removed from the record.

PR #1624, opened 2022-03-30
Suppress empty tools box #1618.

knirirr:
Tools will now not be shown if it's an empty array.


allysonlister:
Thanks very much for this @knirirr it looks great! Here's a screenshot for reference:
ex

I'm curious what the relationship is between the URIs (for the user account and the homepage) at the top of the record (in the summary you get without expanding the record) and the URL that directs to the user account page that is present at the bottom of the expanded section. Are those three different fields, that we can populate with different information?

Thanks!

allysonlister (approved):
I have added and then deleted a tool, and the tool area correctly appears and disappears. Thank you!

ramGranell (approved):

Issue #1620, opened 2022-03-30
Cannot remove a related database on a specific record


delphinedauga:
**Describe the bug**
When I remove the "related_to" Catalogue of Life from the GBIF record (https://fairsharing.org/2163), and I save, nothing happens. I have no error message, it is just that the relation doesn't disappear. Allyson has the same behavior.

**To Reproduce**
1. Go to 'https://fairsharing.org/2163/edit' / Edit Relationships
2. Search ' Catalogue of Life' (or browse the list to find it)
3. Remove the 'Related to'
4. Save and exit
5. Go to the record and see that the relation still exists

**Expected behavior**
The relation should disappear

**Screenshots**
![Mar-30-2022 16-16-35](https://user-images.githubusercontent.com/59688426/160856428-e1e3e81d-0fd3-4840-a6d2-c50bbe110fce.gif)

**Operating system and browser version**
macOS BigSur Version 11.0.1 - Chrome Version 99.0.4844.74 (Build officiel) (x86_64)

knirirr:
This is caused by having the same record linked in more than one different way.

PR #1626, opened 2022-03-31
Duplicate records with different links can now be deleted. #1620

knirirr:
This was affecting any record where there there were multiple associations with the same record ID but different association labels. 

I found a record (3000) with no associations and used that to test, by adding two links to the same  other record and then deleting one.


delphinedauga (approved):
That's fine ! Created 2 different relations for a same record, deleted 1, and it works. Thanks !

ramGranell (approved):

PR #1632, opened 2022-04-04
Reformatted layout of deprecations 1629.

### LINKED ISSUE #1629: Display of chained deprecation links
**Describe the bug**
Chained deprecation links display oddly

**To Reproduce**
1. Go to https://deploy-preview-1622--fairsharing.netlify.app/1195
2. See error

**Expected behavior**
The data model allows this, and in terms of curation it might happen, that record A is deprecated and superceded by B, and B itself is deprecated and superceded by C. This is a chain of deprecations. We don't have it in production yet, but it is a feasible use case.

When testing a different ticket, I discovered that this situation is not handled nicely by the UI. I don't mind if the solution is to keep the entire chain in one row, or to have different rows for each statement, but please choose a solution that will a) read cleanly, and b) scale nicely. After all, in theory we could also have multiple records that are superceded by C, or multiple records in any of the A B or C positions (or indeed a longer chain, though this is unlikely).

**Screenshots**
Both big and small window screenshots:
narrow
wide

Firefox / Ubuntu

knirirr:
I've changed the layout of these elements so they're readable. 
There's not much other option here, given the limits of my html skills.


ramGranell (approved):
I found the information about the related deprecated records readable and clear as the example suggested in the ticket https://deploy-preview-1632--fairsharing.netlify.app/1195

allysonlister (approved):
Looks great - thank you!

PR #1633, opened 2022-04-05
Add display for new notes field

knirirr:
This relates to https://github.com/FAIRsharing/FAIRsharing-API/issues/483

The new notes field will appear in the edit interface, which is populated automatically form the schema, but not on the record page as the component is hard-coded and has various special cases. This modification causes the notes field to appear when added. 

N.B. these fields only apply to database records, e.g. https://deploy-preview-1633--fairsharing.netlify.app/1549


ramGranell (approved):
This field looks fine.

allysonlister (approved):
Thank you this is great!

Issue #1639, opened 2022-04-07
Zenodo DOI call


knirirr:
There's a ticket in the server repo. describing a request for the edit interface to also check Zenodo for publications, using a DOI provided by the user. 

https://github.com/FAIRsharing/FAIRsharing-API/issues/473

That's all handled client-side in the application. Specifically, here:

https://github.com/FAIRsharing/fairsharing.github.io/blob/master/src/lib/Client/ExternalClients.js

This file could be modified to trigger both queries. It may well be necessary to modify this part as well:

https://github.com/FAIRsharing/fairsharing.github.io/blob/master/src/components/Editor/EditPublications.vue#L428-L458

...in order to handle the returned values.

ramGranell:
I think there are two possibilities to check:
1) looking directly for the DOI of the publication (performing a query to the DB over the DOI field) e.g. https://zenodo.org/search?page=1&size=20&q=doi:%20%2210.4122%5C%2Fdtu:229%22 
2) retrieving  the record using the Zenodo URL https://zenodo.org/record/3537300  

I suppose that we are looking for the first.

knirirr:
Yes, it's the first of these.

PR #1642, opened 2022-04-12
zenodo search of publications

ramGranell:
This is related to #1639

The logical is the following: 
1) look in https://doi.org/
2) if not found look in Zenodo with two different cases
   2.a) The text to look for contains the string "zenodo.X", Then I look for directly to recover the X id, e.g. "zenodo.5786374"
  2.b) The text to look for does NOT contain the string "zenodo.X", Then I look for in Zenodo as a generic search being possible to have more than one hit and only allowing the first one. e.g. "j.appet.2021.105433" (that it is not in doi.org but it appears in Zenodo in other field)

I change the button title putting "Import from DOI / Zenodo"


knirirr (approved):

delphinedauga:
Comment n°1 : 
Current format of Zenodo DOI to enter is /zenodo.XXXXXXX (where X are numbers, e.g. '/zenodo.6461459'). 
But it's a bit confusing because on a Zenodo record, I see this : https://doi.org/10.5281/zenodo.6461459, meaning that the DOI is '/10.5281/zenodo.6461459'. 
Is it possible to accept both format ? 
Capture d’écran 2022-04-14 à 16 48 05

Capture d’écran 2022-04-14 à 16 47 59

delphinedauga:
Comment n°2 :
This should be mainly discussed with @allysonlister: 
Currently we can add a Zenodo dataset record in the publication section (e.g. https://zenodo.org/record/5837408#.YlgvltNBzPY).
Do we want that all Zenodo record types to be entered here (including 'dataset', 'image', 'presentation'...), or just the publication type, or a publication subtype ('article', 'book',...):

Capture d’écran 2022-04-14 à 16 53 07

This leads to the re-definition of a publication: 
Is it a peer-reviewed research result or just a published 'report' that can be disseminated in different formats (book, presentation, report, article, thesis, etc.)? 
It is important to decide this, it does not give the same value to the rubric if it is only peer-reviewed articles. We have already had this discussion, but obviously something is bothering me 😬

The issue of our last discussion was : 
Ally do not consider the publication section as needing to be peer reviewed 
That sounded odd. Pre-prints are OK to add
Delph was agree (but it seems that the concept is not understood 😂): a preprint has a publication format, and we can specify if it's published on a journal or a preprint website.

ramGranell:
> Comment n°1 : Current format of Zenodo DOI to enter is /zenodo.XXXXXXX (where X are numbers, e.g. '/zenodo.6461459'). But it's a bit confusing because on a Zenodo record, I see this : https://doi.org/10.5281/zenodo.6461459, meaning that the DOI is '/10.5281/zenodo.6461459'. Is it possible to accept both format ? Capture d’écran 2022-04-14 à 16 48 05
> 
This has been sorted out with a small modification in the API to deal with DOI (https://github.com/FAIRsharing/FAIRsharing-API/issues/473)  with the format: "XXXXzenodo.YYYY" e.g. https://doi.org/10.5281/zenodo.6461459

allysonlister:
Delphine, Susanna and I all agree that we should allow the "publication" type from zenodo, but not the others at this stage.

Thanks!

ramGranell:
> 

The issue is that I'm using the Zenodo API that employs Elasticsearch (our well-known friend) searching in all the fields of their DB (that's the only behaviour I manage to get working for the search as the documentation was not very clear) and I later show the entry of the search that it is the first (Elastisearch decides the order based on their own criterion). I don't know how to tell the API that do not show me that things that are not publications.

> 

As I explained in the first comment of the PR, there are two different search types in the Zenodo API:
1) If the string ".zenodo" appears in the text to search then I assume it is a zenodo style URL and then I obtain this URL, e.g, the word inserted is "10.5281/zenodo.4514014" or "https://zenodo.org/record/4514014" then I get directly the record: https://zenodo.org/record/4514014  that I don't have control if it is a publication or not (in the same way that a DOI it is possible inserted when looking for in doi.org can be that it is not a publication).

2) If the string ".zenodo" does not appear in the text to search, then I make a generic search using the Elasticsearch engine provided by the Zenodod API and get the first entry that it is returned. Then I can put "&type=publication" to indicate that it should be a "publication" in the search: https://github.com/FAIRsharing/FAIRsharing-API/blob/master/app/controllers/publications_controller.rb#L67

In the case 1) I think the curator needs to check what type of element is added as I do not have real control about it.

delphinedauga:
Can you only use the search type 2 ? It will solve the problem, no ?

allysonlister:
I can see the issue - it seems very frustrating Ramon! Could you contact their helpdesk and explain this issue? Ask them how you can only search publications, and also how you can check if a given zenodo DOI is a publication?

Thanks!

ramGranell:
> I can see the issue - it seems very frustrating Ramon! Could you contact their helpdesk and explain this issue? Ask them how you can only search publications, and also how you can check if a given zenodo DOI is a publication?
> 
> Thanks!

I was thinking that the current implementations should be enough robust for not accepting things that are not a publication.: At the moment when you try to retrieve a thing that it is not a publication the journal field is getting empty and then it should be fine as it cannot be created as this is a mandatory field (try it with 10.5281/zenodo.5396847) This issue can happen also with publications that do not have journal assigned such as books e.g. 10.5281/zenodo.4707282 thesis or others. 

You can look for things e.g.https://zenodo.org/search?page=1&size=20&q=zenodo.5786374&type=publication&subtype=book get the id zenodo.5826266 and put in the "import from DOI / ZENODO" box and you will see that journal is empty and cannot be created.

allysonlister:
OK - ultimately I leave the decision to you, @ramGranell (if @delphinedauga agrees). If you feel this is robust enough as things stand, then please go ahead. Thanks!

Issue #1636, opened 2022-04-06
Updated footer


allysonlister:
We are about to add a new community_curation page and it should go in the footer, however, the far right column in the footer is looking a little long. Therefore please can we update the footer as follows (I'm listing each column in the order they appear, from left to right)? Thanks!

1. Developed by : as it is now
2. Contact and News : as it is now
3. About Us :  How to cite us (link as it is now) - Governance (https://fairsharing.org/communities#governance) - Privacy Policy (link as it is now) - API and data licence (link as it is now) - API Documentation (link as it is now) - Terms of use  (link as it is now)
4. Users and Contributors : Adopters (https://fairsharing.org/communities#adopters) - Curation Programme (https://fairsharing.org/community_curation) - Educational (link as it is now) - Stakeholders (link as it is now) - Activities (https://fairsharing.org/communities#activities)

BTW, Susanna says it is fine for her, but when I look at https://fairsharing.org/communities#activities on Ubuntu/Firefox, this URL (for me) doesn't link to the top of Activities, but partway down. Does it work properly for you all? If it does, no worries, it's just something odd on my browser.

Thanks!

PR #1644, opened 2022-04-19
Rg footer change #1636

ramGranell:
That's related to ticket #1636 to update the footer as suggested.


knirirr (approved):

allysonlister (changes requested):
This looks great! One very small change please. 

Could you change the text in the right-hand column of the footer from "Curation Programme" to "Community Curation" please?

(I realise I'm the one who gave you "Curation Programme", but now that I see it, I think "Community Curation" might suit it better)

Thanks!

allysonlister (approved):

Issue #1656, opened 2022-05-03
"Add organisation" popup should check alternative names for existing organisations


allysonlister:
Currently, when adding a new organisation to  a record, the user is able to search the full names of all existing organisations. However, as we begin to use our new alternative name field, we will need this search field to also be able to search alternative names as well as full names.

example: https://fairsharing.org/organisations/1784

![heart-and-lungh](https://user-images.githubusercontent.com/143586/166461377-2fb662d5-0ad9-4ef6-8fff-274ed2aadc27.png)
![mpi-hlr](https://user-images.githubusercontent.com/143586/166461383-df201510-6299-4de1-8e40-84afc23517fb.png)

PR #1657, opened 2022-05-04
Add alternative names to autocomplete data. #1656

knirirr:
This requests alternative names from the server (previously not available) to allow filtering of them in the client's autocomplete by means of simple text comparison. 

Best to try this after approx. 14:00 today to give time for indexes to be rebuilt.


ramGranell (approved):
The search of organisations works for alternative names such as MPI-HLR

allysonlister (approved):
Great, thank you!

Issue #1675, opened 2022-05-23
Confirmation dialog for Request Ownership


allysonlister:
Please can we add a confirmation dialog for the Actions -> Request Ownership menu item?
![actions](https://user-images.githubusercontent.com/143586/169858643-a574075e-2ea1-4cbf-894b-0369b55a979e.png)

We have lately had a number of accidental claiming requests that wastes our time and their time... So when a user chooses "Request Ownership", please can we pop up a confirmation dialog that says 

"Please confirm that you would like to request ownership of this record. 
We encourage resource developers to claim ownership of their FAIRsharing record; doing so will allows them to make updates and provides them with a visible means of attribution for their efforts. However, please note that for this request to be approved, you **must** be associated with the institutions(s) that develop this resource. Please also see our documentation on [requesting ownership](https://fairsharing.gitbook.io/fairsharing/#claiming-a-record)."
And then buttons for yes and no, please

Thanks!

PR #1677, opened 2022-05-24
Added a confirmation dialog for claiming records #1675.

knirirr:
I've added a dialog along the lines of existing ones.


ramGranell (approved):

allysonlister (changes requested):
Thank you - the functionality is correct, however there are a few things please:
1. I noticed a typo in my text for the popup - please would you replace "doing so will allows them to" with "doing so allows them to" (remove the "will").
2. The size of the text is rather large - could we shrink it a bit?
3. When I shrink the browser window size the popup stays the same size. If this is too difficult then don't worry about it, having slightly smaller text as described in point 2 should suffice.

Thanks!

knirirr:
I can't do anything about point 3 but the others can be addressed (that offending word was on a line wrap and so hard to spot).

allysonlister (approved):
Thank you - looks good!

Issue #1688, opened 2022-06-07
"Extends" relationship for policies - verb usage


allysonlister:
It appears as though the same word, "extends", is used to indicate a relationship between two policies, irrespective of directionality.

Example:
https://fairsharing.org/3408 (Nature Scientific Data - Data Repository Guidance) "extends" https://fairsharing.org/3415 (Springer Nature Research Data Policy Type 4), as defined within its metadata. However, in 3408 the record display provides this sentence: "Springer Nature Research Data Policy Type 4 *extends* Nature Scientific Data - Data Repository Guidance".

![ex](https://user-images.githubusercontent.com/143586/172332430-4ca03782-b3e0-4a8e-8973-433de8d60a92.png)
(https://fairsharing.org/3408)

Context is important here, and what is true in the direction curated is NOT true in the reverse direction.

I see two options (unless you see another?), feel free to do whichever you prefer:

1. Continue only using "extends", but ensure that, when you always mention the source first. In the example above, that would change the sentence to  "Nature Scientific Data - Data Repository Guidance *extends* Springer Nature Research Data Policy Type 4".
2. Use a different phrase (e.g. "is extended by") when the target is mentioned first (as in the example above). This would change the sentence to "Springer Nature Research Data Policy Type 4 *is extended by* Nature Scientific Data - Data Repository Guidance".

Thanks!

knirirr:
There's a comment in the relevant PR about the effect this change will have: https://github.com/FAIRsharing/fairsharing.github.io/pull/1689#issue-1263342805

PR #1689, opened 2022-06-07
Another change to format of related records etc. #1688

knirirr:
I've modified the utility file which formats the records for display, and this has corrected the display issue mentioned in #1688. Please note, though, that it will affect everything else as well. So now, for example, one might see database relationships going both ways on a record. As an example, here are a couple from https://deploy-preview-1689--fairsharing.netlify.app/FAIRsharing.7g1bzj 

Screenshot_20220607_145128


allysonlister:
I'll look at it now - I should say, however, that we already saw this behaviour for databases (e.g. https://fairsharing.org/FAIRsharing.7g1bzj)

allysonlister (approved):
Thank you, looks good!

ramGranell (approved):

PR #1695, opened 2022-06-09
Changed method of watching/unwatching records.

knirirr:
This should fix the problems for the user by using an entirely different method of changing which records are being watched.

https://github.com/FAIRsharing/FAIRsharing-API/issues/531


ramGranell (approved):

allysonlister (approved):
I watched 8 additional records, all were displayed properly in my user page. I unwatched 1, and it disappeared correctly too. Thanks!

knirirr:
Thanks. I'll deploy this and let the user know. 
With luck this will work around the problem.

Issue #1509, opened 2022-02-03
Login session expires randomly


stain:
**Describe the bug**
After logging in with ORCID, navigating the site will easily cause the session to become logged out.

**To Reproduce**
1. Log in with ORCID
2. Navigate to an entry - remains logged in
3. Click on a DOI of a random entry
4. After coming back to fairsharing.org you are now logged out, as indicated by top-right corner.

You can alternatively try to use the Back button to get to an entry you visited before logging in, no matter of Refresh will get it to log in.

This means I am forced to log in every time I try to edit an entry.

Navigating Fairsharing in a second tab will also cause it to be logged out.

This may be the cause of "Access denied" in https://github.com/FAIRsharing/fairsharing.github.io/issues/1488#issuecomment-1028861061

**Operating system and browser version**
Win 10 - Firefox 96.0.3

knirirr:
Thanks for the report. This is unrelated to #1488 - server side code has been modified to address that. 
We'll investigate this issue.

knirirr:
This is proving very hard to reproduce.
Would you mind providing an example of a DOI upon which you click which can consistently cause this problem?
Also, do you find the same problem if you log in directly rather than going via ORCID?

knirirr:
This can be re-opened if any means to reproduce it can be reported.

PR #1701, opened 2022-06-10
Millisecond correction for login. #1509.

knirirr:
To check this, please do the following:

1. On localhost:8080, see if the problem persists by logging in, immediately refreshing the page and seeing if now logged out.
2. If (1), then run this branch locally and see if the problem disappears.

It may also be possible to try this using Netlify.


ramGranell (approved):
All seems to work properly locally

Issue #889, opened 2021-04-15
Error message styling


knirirr:
When editing in particular, error messages from the server show as the raw JSON, e.g. 

`{ "message": "Record modification is currently disabled." }`

PR #1712, opened 2022-06-15
Some styling for error messages #889.

knirirr:
This addresses #889 by using the same technique as used for logs; the method I used for login errors messes with important aspects of the format of error messages when editing records. 
As it is difficult to force messages to appear, here are a couple of examples I made earlier:
Screenshot 2022-06-15 at 16 34 42
Screenshot 2022-06-15 at 16 32 56

I am tempted to apply this to the login error messages as well.

### DIFF
--- a/src/components/Editor/Alerts.vue
+++ b/src/components/Editor/Alerts.vue
@@ -8,8 +8,16 @@
         :type="message.type()"
         class="mb-0"
       >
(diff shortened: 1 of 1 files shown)


allysonlister (approved):
thank you this does look cleaner!

ramGranell (approved):

PR #1714, opened 2022-06-16
Opt in to orcid

### LINKED ISSUE #1611: Additional switch for user profile editing
See this document for context: 

https://app.gitbook.com/o/-MchtBAOWXh2gO--zz7m/s/-MkSKpWh9NA-iywIqhP_/please-can-we/pushing-attribution-to-orcid#specification-in-fairsharing

> When editing a profile: Toggle to be added (default yes if the user has provided their ORCID ID, and automatically flips to yes if an ORCID ID is added at a later date) called "allow the crediting of my FAIRsharing curation to my ORCID profile".

knirirr:
This deals with #1611. Please note that it's not possible to have the ORCID opt-in toggle turn on visibly on the page whilst a user types in their ORCID ID as we have disabled ORCID ID addition/editing and users may only add one by authenticating with ORCID. 

So, once they so authenticate, the preference will be set to "true". They may later come to their profile page and set it to "false" if they so desire. 

There's a corresponding PR for the server at https://github.com/FAIRsharing/FAIRsharing-API/pull/541

N.B. using ORCID other than in production involves running this application at localhost:8080 and using an ORCID sandbox account.


allysonlister:
Does using localhost / sandbox ORCID mean I don't test it, just see if, from the comments, it makes sense?

knirirr:
> Does using localhost / sandbox ORCID mean I don't test it, just see if, from the comments, it makes sense?

Your choices are:
1. Run everything locally, sign up for the ORCID sandbox and go through the whole thing, or...
2. Glance at the code (if you're a developer) and your profile edit page (and perhaps someone else's) just in case anything leaps out at you.

Either will do here; I'm sure it's fine (hence the PR ;-) but a second glance is always useful.

ramGranell (approved):

allysonlister (approved):
Thanks for the advice, I've done as you suggest, all looks OK!

PR #1719, opened 2022-06-17
Additional info change count 1718

### LINKED ISSUE #1717: Record delete button
This offers the options of "cancel" and "claim" instead of "cancel" and "delete".

### LINKED ISSUE #1718: Changes to additional information
Adding or removing an element (e.g. an access point) will show a count of changes on the relevant tab, as expected. But, editing an existing element will not.

knirirr:
This deals with both #1717 and #1718. 
When looking at the count of changes it should be counting the number of fields which have changed since the record was loaded. This is done by a JSON deep diff.


ramGranell (approved):
The counting seems to work fine.

PR #1724, opened 2022-06-21
Fixed bug described in #1723.

### LINKED ISSUE #1723: Pagination not shown correctly at the bottom of search results
A user reported that the pagination controls at the top of the page perform as expected but those at the bottom do not. For example, if one were to advance to page 2 or 3 at the top this would not be reflected on the bottom. 
However, the bottom control does appear to change the page correctly even if it does not display it.

knirirr:
This now behaves as it should.
Also, a mysterious extra pagination object has been removed.


ramGranell (approved):
Pagination seems to work properly.

PR #1733, opened 2022-06-27
Added hack to work around #1732.

### LINKED ISSUE #1732: Save buttons in General Information greyed out
**Describe the bug**
 "General Information" tab in https://fairsharing.org/1567/edit has odd behaviour in the save buttons. When I load it, the "Save and Continue" as well as "Save and Exit" buttons are greyed out, though the other tabs' buttons look OK

**To Reproduce**
1. Go to https://fairsharing.org/1567/edit
2. scroll down to save buttons
3. Buttons are greyed out. 

**Expected behavior**
Save buttons should be green.

knirirr:
Now, on loading https://deploy-preview-1733--fairsharing.netlify.app/1567/edit you should see exactly why the form isn't valid.
Please try editing another record or two just in case I've broken something, but from what I can tell all is well.

### DIFF
--- a/src/components/Editor/GeneralInformation/BaseFields.vue
+++ b/src/components/Editor/GeneralInformation/BaseFields.vue
@@ -1,345 +1,62 @@
 <template>
-  <v-row>
-    <v-col
-      xl="4"
-      lg="6"
-      md="12"
-      sm="12"
-      xs="12"
-      cols="12"
-    >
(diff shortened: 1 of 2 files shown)


ramGranell (approved):
I tried 6-7 random records and all seems to work properly.

allysonlister (approved):
Thank you for this! All good for me too

PR #1743, opened 2022-07-04
Record history can now be loaded. #1742

### LINKED ISSUE #1742: Direct link to edit history
Currently, a direct link to a record's changes is not available, and users don't always see instructions on how to find that information when they receive emails from us. 

Perhaps a parameter could be added to a record's URL which would trigger the action to load record history, if that URL were loaded.

knirirr:
To see this in action, add `?history=show` to any record URL. For example:

https://deploy-preview-1743--fairsharing.netlify.app/FAIRsharing.9kahy4?history=show

Users are supposed to be logged in for this to work. However, clicking that link whilst not logged in will cause the history to be displayed anyway. If/when this change is approved I'll make the necessary changes to the email text. 

The tests are a nuisance and if anyone has any idea about the problem, please let me know.


ramGranell (approved):
Updates are nicely displayed. The issue that we are going to find is that most of the users are not going to understand the meaning of these updates. In some place (gitbook?) we will need to indicate what the different fields and operations that are shown there mean.

In addition, you said that you need to be login to check the history, but is it possible for a login user to see the history of a record that is not a maintainer?

Sorry, I cannot help you with the  js tests.

knirirr:
Yes, logged-in users are supposed to be able to see the record history, no matter whether they maintain the record or not. 

Forgot to add: Any change to the display of edits might require a bit of work, and would be a topic for discussion and then a separate ticket/PR.

allysonlister (approved):
Looks great, thanks!

PR #1766, opened 2022-07-25
Added a status icon for records in the edit form.  #1763

### LINKED ISSUE #1763: Display record status within relationship tab when editing records
The relationship tab looks like this, with two sections, one showing the possible records to link to the current record, the other showing the records **already** linked to the current record:

Please can we update this tab such that the status of each displayed record is also shown? In this way, it is easy for users to see if they're about to add a deprecated, uncertain, in dev or ready resource to their own record. In the [example below](https://fairsharing.org/4258/edit), it was only when I clicked the preview icon (which I often don't need to do) that I noticed the FDGC record was deprecated.

We already highlight rows as shown in the first image above, so I wouldn't suggest doing this with highlighting. Although I'm happy to hear other ideas, the little circle icons that are overlayed on the record type might be re-used here and displayed in each row of the search result, perhaps next to the "eye" icon for previewing the record e.g.

Thanks!

knirirr:
There's an existing component which shows a record's type/registry overlaid with a status icon, and it already contains a switch to display only the status. I have used this in the position requested for records in the left column as there's already a type/registry icon to the left. I tried fitting the status icon on the left, but had some difficulty.
Status only appears on the left column (i.e. records not-yet-added to relations) as the data which are used for the right-hand column don't include any status information.
It wasn't clear to me from the ticket if the icons were being requested for both columns; if not then this PR can be reviewed as it stands. But, if they are required, I'll have to open another ticket for changes to the server so that it can supply record status in the relevant part of the data.


ramGranell (approved):
It seems to work correctly this new record status in the edit symbol system. My only comment is that when your mouse is over this new status symbol it says "Ready", "Uncertain" that is very useful. However, when I put the mouse over the symbols next to it (the eye and arrow) I was expecting to find an information sign saying "View" and "Add" in the same style as it happened with the status icon. That's only a comment that is not related with the original ticket.

allysonlister (approved):
Thank you, this looks great. You both also raise valid points. Addressing these:

1. I think we should have a separate ticket to add hover-over text for the preview and add, as Ramon suggests
2. You're right @knirirr  I wasn't clear in this ticket if the status should be on both tabs. Let's complete this one as-is, but then please could you create another ticket to add the same information on the right-hand side.

Thanks!
Allyson

prakhyatox (approved):

PR #1787, opened 2022-08-18
Fixed rule to account for empty array #1786.

### LINKED ISSUE #1786: New records can be submitted without a type/registry
If the registry/type of a new record (on /create) is not specified then the form can still be submitted; the record will not be created as the server will return 422 due to the missing field.
A validation rule on this field would be necessary to prevent submission. 
There also appears to be a bug in that the "create record" button may be clicked even if there are required fields which are unfilled.

knirirr:
Supplying a "type" is now compulsory when creating a new record.
This was caused by Javascript not considering an empty object to be false.


prakhyatox (approved):

knirirr:
This went into master by mistake; I'll merge it into dev also.

PR #1797, opened 2022-08-25
Changed some spellings.

knirirr:
This brings some static pages in line with spellings elsewhere.


ramGranell (approved):
Spelling changes look more Britishly correct!

PR #1804, opened 2022-08-30
Hidden collection records

### LINKED ISSUE #1802: User authentication for search
Users' tokens aren't passed to search methods, i.e. searching all records and searching a collection. Therefore, records which should be revealed to admins, aren't. 
This should have been added initially but has been overlooked.

knirirr:
This PR deals with sending users' authentication information to the server when searching, either by using the usual search page or the parts of it which were later adapted for collection pages. The result is that hidden records should now be visible to curators (and above). 

This is discussed at: https://pyroeis.oerc.ox.ac.uk/#narrow/stream/8-FAIRsharing-Rebuild/topic/Record.20Relations/near/41596

If no Netlify build appears then it will have to be run locally, but using dev-api.fairsharing.org as I've copied the latest data there so that the record discussed above can be looked at, in addition to looking at the code.


allysonlister (approved):
I have checked a few records, including the one in question, and all looks good. I have not checked the code. Thanks!

ramGranell (approved):
It seems to work properly. In the code I saw that you need to pass the user token as parameter to the query to get both the collection and the search data.

knirirr:
> It seems to work properly. In the code I saw that you need to pass the user token as parameter to the query to get both the collection and the search data.

Indeed - the server was always set up to handle this, but adding the token to these parts of the client was never done, which I had not realised.

Issue #1753, opened 2022-07-19
Add Summary Counts graphic to homepage


allysonlister:
Please can we have a set of summary "boxes" on the FAIRsharing homepage to showcase some basic statistics about FAIRsharing? Underneath the boxes could be a link to our statistics page (e.g. with the text "full statistics" or similar). 

## Location
This should go in the dark blue section of the page, in between the general search and the subject browser. I'm not 100% convinced it will fit there, so if you have different ideas I'd be happy to hear them. To help it fit, please could you put the "Browse by Subject" button below the subject browser logo, and perhaps the size of the white search box could be slightly shrunk. If it really doesn't seem to fit there, then try it in its own area directly above the dark blue section please.

## What Summaries should be included?
| Box text | how to get the value for NNN | comments|
|--------|----------|-------|
|NNN Community contributors| number of users with "curator" role + number of users who maintain at least one record of any record status|needs separate ticket for REST endpoint. |
|NNN Resources described| number of standards + databases + policies| Note that collections are not included in this count, but all record statuses should be included.
|NNN Views since 2011|from google analytics |The "since 2011" should be in smaller footnote text below the main box text. Unknown if https://developers.google.com/analytics/devguides/reporting will provide this. Additional comments on this item in the next section below.|

## Notes on "Views" stats

The questions I have regarding the Views stats are shown here. Mostly, I'm looking for the best way to make a few choices, bearing in mind our current audience overview on google analytics has the way the page visits are counted changed since we implemented the new site? We've gone from 4000+ users in a month to much less, and I'm sure that's not right... If necessary, please make another ticket to investigate this.

Please can I have developers' opinions on 

1. unique vs non-unique counts of site visits? 
2. ideally, this would be over all time, since biosharing.org started. I realize it may or may not be possible to get this info from the biosharing google analytics page, but once we have the number it can just be permanently added to the running total we get from google analytics for fairsharing.org. If only the fairsharing.org stats can be used, then we need to update the box text to "Views since 2018"
3. What is the right terminology here? I've called the box text "NNN views since 2011", but should we use views, or another word such as visits/users/something else?

knirirr:
The stats are now available at https://api.fairsharing.org/homepage_stats, though the values won't be correct until Elasticsearch has been reindexed.

allysonlister:
I'm not sure if my input is required at this stage, but the stats look good to me, thanks!

PR #1808, opened 2022-08-31
Statistics count 1753

prakhyatox:
Ticket
https://github.com/FAIRsharing/fairsharing.github.io/issues/1753

WDID:
Added new StatisticsBlock.vue component to show the analytics data

image


prakhyatox:
@allysonlister Please verify the design of the statistics block (via https://deploy-preview-1808--fairsharing.netlify.app/) and suggest changes if any

knirirr (approved):
I'm happy with this, but given the nature of this change then some curator feedback would be necessary as well.

allysonlister:
Thank you this is looking great! After Susanna and I reviewed it, we feel that we need a little more prominence to this work. So, my apologies, but please could you do the following:

1. Move these three items to their own area, after the final carousel and before the "View all" button.
2. give each item their own box positioned one after the other horizontally
3. Make the numbers the biggest, and the text below them just a little smaller.

Something a bit like this from RDMkit, collaborators of ours:
image

Each box would say:

Box 1:
1096
**Contributors**
Our community of resource developers and curators

Box 2:
3706
**Records**
Describing standards, databases and policies for humans and machines

Box 3:
192k
**Views**
Community usage since 2018

Issue #1817, opened 2022-09-02
Display Attributes and Conditions data


prakhyatox:
**Describe the bug**
Rows like Data Curation, Data Access Condition, Data Deposition Condition, Resource sustainability etc. within Attributes and Conditions section should be displayed if they have the data otherwise respective rows should be hidden

**To Reproduce**
1. Go to 'https://fairsharing.org/dryad'
2. Scroll to ```Attributes and Conditions``` block within ```DATA PROCESSES AND CONDITIONS```.
3. If any row like ```Resource sustainability``` does not have any data, that row should not be displayed
 

**Expected behavior**
<img width="675" alt="image" src="https://user-images.githubusercontent.com/109139583/188123662-ee4b3464-dd89-43da-848a-657f9216ef5a.png">

**Actual behavior**
<img width="696" alt="image" src="https://user-images.githubusercontent.com/109139583/188123235-b505269b-aae6-46f8-88b1-6ec65425a463.png">

PR #1818, opened 2022-09-02
Updating display 

prakhyatox:
Ticket: https://github.com/FAIRsharing/fairsharing.github.io/issues/1817

WDID:
1) If any attribute does not have a data, it will not be displayed in the block e.g in screen shot below ```Resource Sustainability``` does not have any data so field is not displayed
2) If any attribute have only name/type like ```Data Preservation Policy``` in the screenshot and no url, it will be displayed a normal text instead of hyperlink 

image


allysonlister (approved):
Thank you!

knirirr (approved):

PR #1820, opened 2022-09-02
Added a tooltip on ROR button mouseover.

knirirr:
I've added a tooltip on the ROR import button to clarify how it works.


allysonlister (approved):
Thanks - this makes a lot of sense!

PR #1826, opened 2022-09-09
Fixes #1825.

### LINKED ISSUE #1825: Validation of URLs with long domains
On the record editing/curation forms the URL validator doesn't like longer domains e.g. .science; presumably it expects only three characters there. This has affected one record of which I'm aware (https://fairsharing.org/4385).

knirirr:
This allows up to 8 characters at the end of a URL, which will allow .science as in #1825.


ramGranell (approved):
The change works, but maybe it will be necessary to increase the length longer in the future, e.g. 10 or 12?

knirirr:
That thought had occurred to me, though I think longer domains are rare. I'll keep an eye on it and will modify further if necessary.

PR #1834, opened 2022-09-22
Added offsite redirect for ontology terms. #1833

### LINKED ISSUE #1833: Resolvable SRAO IRIs for individual terms
An update to previous work (e.g. #1741 please)

A user has requested resolvability of our SRAO IRIs - this should be possible. Most of the IRIs in SRAO come from other ontologies and therefore we don't need to worry about their resolvability, however we should have redirections set up for our own IRIs.

Example:

http://www.fairsharing.org/ontology/subject/SRAO_0000307

Could resolve to https://www.ebi.ac.uk/ols/ontologies/srao/terms?iri=http%3A%2F%2Fwww.fairsharing.org%2Fontology%2Fsubject%2FSRAO_0000307 which would be useful.

You can see from these examples how to construct URLs, so could a rule be set up such that any URL with the pattern http://www.fairsharing.org/ontology/subject/SRAO_* resolves to the appropriate EBI URL as shown above?

Thanks!
Allyson

knirirr:
This directs individual SRAO queries offsite as per #1833.

### DIFF
--- a/src/router/index.js
+++ b/src/router/index.js
@@ -178,6 +178,16 @@ let routes = [
             }
         }
     },
+    /* See: https://github.com/FAIRsharing/fairsharing.github.io/issues/1833 */
+    {
+        name: 'srao_term',
+        path: '/ontology/subject/:id',
+        redirect: to => {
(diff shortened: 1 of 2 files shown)


allysonlister (approved):
this works great, thank you!

prakhyatox (approved):

Issue #1845, opened 2022-09-28
Empty Processes section in some records


allysonlister:
As with #1839, it appears that the Processes section also appears empty for some reason here:

https://fairsharing.org/4417
![image](https://user-images.githubusercontent.com/143586/192741584-4a3cc946-cf6c-4760-a2c1-c920fa69342f.png)

I need to update and approve this record, so I hope you can proceed with this anyway. It might be worth investigating if any of the other sections within Data Processes and Conditions also have this issue but is as yet undetected.

PR #1847, opened 2022-09-28
hide-processes-section-when-empty

prakhyatox:
Ticket: https://github.com/FAIRsharing/fairsharing.github.io/issues/1845

WDID: 
hide Processes Section when there is not data entry for that field

image


allysonlister (approved):
I've tried a few random records but all have looked OK - I tried to test with 4417 but it's not in the deployment here. Approving based on screenshot and a quick review of the diff. 

Thank you!

knirirr (approved):

prakhyatox:
If you can edit any record in netlify URL, if it is safe to delete the data (assuming it wont impact the dev after merge) of in processes and condition section and verify then processes section should be hidden

image

Issue #1848, opened 2022-10-03
Data processes and condition section is hidden


prakhyatox:
**Describe the bug**
Data processes and condition section is hidden even when ```Attributes and Conditions``` section has some data

**To Reproduce**
1. Go to 'https://fairsharing.org/4337'
2. Scroll down to see 'section DATA PROCESSES AND CONDITIONS'
3. No ```DATA PROCESSES AND CONDITIONS``` section is present

**Actual Behavior**
![image](https://user-images.githubusercontent.com/109139583/193590513-bc5dfcf5-62c9-49dd-a5d4-41a25c6ffa1d.png)

**Expected behavior**
![image](https://user-images.githubusercontent.com/109139583/193590577-8223149a-5600-468c-b7dc-c548452f49f9.png)

PR #1849, opened 2022-10-03
show data-processes-and-condition-block

prakhyatox:
Ticket: https://github.com/FAIRsharing/fairsharing.github.io/issues/1848

WDID: 
Show ``Data Processes And Conditions`` block

image


allysonlister (approved):
I'm checking this only on my phone, so only looking at the screenshot. As long as there is a second reviewer to check the code, I'm happy for this to go forward and long as a few other random records in the database registry have been double checked?

Thanks!

ramGranell (approved):
I checked few records and it seems to work properly

PR #1863, opened 2022-10-12
Changed preview behaviour to fix #1782

### LINKED ISSUE #1782: Possible editing of multiple records during a single edit session
**Describe the bug**
@allysonlister reported a bug in our internal system that couldn't be reproduced by our technical team - yet (they are still working on it). However, @lnanderson, one of our community curators, has now reported similar behaviour and we're hoping that this extra information might help @knirirr et al. figure out what's happening.

Sometimes, when one record is being editing in the UI, one or more of the changes are saved on different records.

Apologies, this one looks tricky. Lindsey and I, in trying to figure out what we are both doing, are wondering if it has something to do with toggling tabs while performing multiple edits. We both often toggle back and forth among all the tabs, editing bits and pieces and saving intermittently. Could it be that when we toggle, sometimes we accidentally move to another record when you move tabs, just for that tab? One "next to it" in some way (alphabetically? via relationships? ids?) ?

**To Reproduce**

As I said, it is not easily reproducible, and @knirirr and the team are trying to reproduce it. So instead of steps to reproduce, we'll provide some examples here.

## Example 1: from Lindsey

@allysonlister queried an edit Lindsey performed:
> It actually seems as though the Kbase organisations were accidentally added to WholeCellKB's record (https://fairsharing.org/1701)? Please would you check, and update as required?

Lindsey replied:

> for record https://fairsharing.org/1701 (WholeCellKB). This is a bug I noticed a while back but since it keeps occurring and now is popping up elsewhere (based on your query of it), I figured we can elaborate on it here. This bug has popped up more than once for me, specifically for module entries in the “Licenses and Support Links” record field. During active/live curation, when I have saved input entries during live edits of my own, after  toggling back and forth across the other fields in the record edit, all of a sudden it either deletes and replaces my entries with other records that sharing similar naming schemas (like WholeCellKB and KBase, both have similar acronyms in the name). Like I mentioned this only occurs in this specific record field tab, and sometimes will auto backfill information from other records that I have not edited or have been associated with curating and replaces the resources I have added where I have had to go back and redo them. I have noticed if i back arrow in the browser and then refresh, it repopulates the information. It may be worth having a developer look at. If my explanation isn’t clear please let me know, it is a sort of complicated error to explain.

## Example 2: From @allysonlister 

> I have just been editing https://fairsharing.org/888 (Sickle Cell Disease Ontology (SCDO)) please see the record history if you wish, but I modified most every section of the record. One of the things I added was adding https://fairsharing.org/organisations/1207 as an organisation.
> I then approved this update in the dashboard. Then, just below 888 in the dashboard, I noticed https://fairsharing.org/1275 (Infectious Disease Ontology Core (IDO)) had an update pending for me. However, I promise I did NOT update that record (it is, however, related to 888 via its relationships).
> But for some reason, 1275's record history has me adding https://fairsharing.org/organisations/1207 to this record rather than 888. Further, although all of the other edits for 888 seem to be saved, it does NOT have https://fairsharing.org/organisations/1207.

knirirr:
The fix here was easy - instead of previewing a record when adding relations the button opens a link to that record in a new browser tab. The preview box was overriding the value of currentRecord['fairsharingRecord'] in the edit form, and this can be observed by watching the record ID at the top of the edit form as you click on "view record" eye icons. In this branch it will not change; in dev/master, it will.

### DIFF
--- a/src/components/Editor/EditRelationships.vue
+++ b/src/components/Editor/EditRelationships.vue
@@ -110,7 +110,7 @@
                             icon
                             class="blue white--text mr-2"
                             v-on="on"
-                            @click="showPreviewOverlay(record)"
+                            @click="viewRecord(record)"
                           >
                             <v-icon small>
                               fas fa-eye
@@ -240,7 +240,7 @@
                           icon
                           class="blue white--text mr-2"
                           v-on="on"
-                          @click="showPreviewOverlay(association.linkedRecord)"
+                          @click="viewRecord(association.linkedRecord)"
                         >
                           <v-icon small>
                             fas fa-eye
@@ -417,22 +417,6 @@
       </v-container>
     </v-dialog>
 
-    <!-- PREVIEW RECORD -->
-    <v-dialog v-model="showPreview">
-      <v-btn
-        fab
-        small
-        class="grey--text absolute"
-        @click="showPreview = false"
-      >
-        <v-icon>fa-times</v-icon>
-      </v-btn>
-
-      <v-card>
-        <Record :target="targetPreview" />
-      </v-card>
-    </v-dialog>
-
     <!-- attempt to add duplicate relationship -->
     <v-snackbar
       v-model="duplicateRelationship"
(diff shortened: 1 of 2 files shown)


prakhyatox (approved):

allysonlister (approved):
I can't get the id to change as I could before - thank you for this!

ramGranell (approved):

PR #1866, opened 2022-10-13
Hide tooltips if description is missing for dynamic edit form elements.

knirirr:
This relates to https://github.com/FAIRsharing/FAIRsharing-API/issues/593, and hides tooltip icons if there's no description associated with a field.


prakhyatox (approved):

allysonlister:
Ideally, I would like the justification for the fields without "?" to match the other rows that have "?" so that it all lines up nicely. Would this be not too much trouble? Thanks!

knirirr:
I don't think I'm able to do that, sorry.

allysonlister (approved):
No worries. This shouldn't ever happen - we should have tooltips for everything. I am happy to proceed as such

Issue #1865, opened 2022-10-12
Redirections for old URLs


allysonlister:
Hi all!

There seem to be a few convenience URLs that used to be used on the old site that aren't redirecting to their equivalents. Please could you create these redirections?

old URL | Redirects to
---|---
https://fairsharing.org/standards/reporting_guidelines|https://fairsharing.org/search?fairsharingRegistry=Standard&recordType=reporting_guideline&page=1
https://fairsharing.org/standards/reporting_guideline|https://fairsharing.org/search?fairsharingRegistry=Standard&recordType=reporting_guideline&page=1
As the two rows above, but for every record type across standards, databases and data policies| "
https://fairsharing.org/recommendations|https://fairsharing.org/search?isRecommended=true&page=1&searchAnd=false&fairsharingRegistry=database,standard
https://fairsharing.org/license|https://fairsharing.org/licence
https://fairsharing.org/communities/activities|https://fairsharing.org/communities#activities
https://fairsharing.org/HUPOPSI|https://fairsharing.org/3514

These were taken from various publications (RDA outputs, FAIRsharing publication etc) and checked and are currently 404.

Thanks!

knirirr:
Re. that last one, this is available: https://fairsharing.org/HUPO%20Proteomics%20Standards%20Initiative
As is: https://fairsharing.org/HUPO-PSI

Where is the offending link?

allysonlister:
In one of the manuscripts the URL doesn't have a dash between HUPO and PSI and that gives a 404... I don't want to change the record to accommodate this as HUPO-PSI is correct, so having the redirect would be useful :) thanks!

PR #1875, opened 2022-10-17
Redirection hackery 1865

knirirr:
Please have a look at the changes in the code to check that the URLs are exactly as required;  #1865 doesn't specify all the details so I might not have interpreted things as intended (N.B. the redirections are all covered in the unit tests).


allysonlister (approved):
https://deploy-preview-1875--fairsharing.netlify.app/communities/activities redirects to https://deploy-preview-1875--fairsharing.netlify.app/communities#activities, but this URL doesn't work on deploy though it does in production https://fairsharing.org/communities#activities - not sure why...

Other than that, it looks great, thank you!

ramGranell (approved):

Issue #1858, opened 2022-10-11
Addition of a "non-exhaustive / exhaustive" toggle for the licences field


allysonlister:
# Background

([Background](https://fairsharing.gitbook.io/fairsharing/record-sections-and-fields/licences-and-support-links/licences) on licences) 

Based on user requests, and my past curation experience, there are quite a few resources that say e.g. "All CC licences are allowed, but we recommend the use of CC-BY or CC0". There is no practical way to curate all of these lists I'm going to list all CC licences, but after discussion with @delphinedauga and @ramGranell we feel it is worth saying whether or not the list of licences in our record is exhaustive or non-exhaustive.

This will allow maintainers to say essentially either "These are just some examples of the licences we allow" versus "You must use one of these licences". There are lots of resources that describe one or more licences explictly but then say "and also any 'open' licence" will do. I think this difference might be important to capture.

# Enhancement

Addition of a yes/no toggle to be connected to the licence field in a record. The toggle text should be: 

"As described in our [licence documentation](https://fairsharing.gitbook.io/fairsharing/record-sections-and-fields/licences-and-support-links/licences), you may choose to provide a complete list of all applicable licences, or a subset of licences that are recommended by your or most commonly used by your users. Please let us know if the licences you have provided are an exhaustive/complete list, or are non-exhaustive/partial list of licenses." (default non-exhaustive). 

I will update our gitbook docs once this is completely. Thanks!

I realise this may require additional tickets for the backend but I thought I would start here and you could choose what other tickets to make.

PR #1890, opened 2022-10-21
Exhaustive licences 1858

knirirr:
This adds a new field as requested in #1858, with the addition of some text on record pages to indicate the status of this field for each record.


prakhyatox (approved):

allysonlister (approved):
Thank you, this is looking great!

Issue #1880, opened 2022-10-18
Github actions Node version update


knirirr:
The details can be found in this post: https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/

terazus:
If i remember correctly the site builds of both 16 and 18.
I suggest you try out nvm (https://github.com/nvm-sh/nvm) to handle mutliple node install.

knirirr:
I wasn't able to update all actions to 16, but I updated the docs action from 10 to 14; apparently, warnings are being shown for 12 and below so this should keep us going for a while.

PR #1896, opened 2022-10-27
Changed actions to use node 16. #1880

knirirr:
As per #1880, this PR should be OK if all the tests build/pass and the documentation builds.


knirirr:
All I've done is changed the documentation build version of node from 10 to 14; Github is showing warnings for 12 or below. Trying 16 caused problems with other packages.

prakhyatox (approved):

Issue #1922, opened 2022-11-24
December 1: Remove any people from our_curators without the role of "curator"


allysonlister:
On December 1, any person accepted to the community curation programme who has not yet received a role of "curator" has not onboarded with the programme, and therefore should be removed from /our_curators. They are more than welcome to reapply at a later date - we always have many calls on our time.

Therefore, on the 1st please check the roles of all of the people on /our_curators, removing them from the page. Before performing the removal, please send me a list of those to be removed so I can confirm.

Thanks!

allysonlister:
How is this looking? Do you have the list of users from our_curators who are not yet in the curator role please?

allysonlister:
I have checked this list. Please note that I have now moved Lars from "user" to "curator". The rest can now be removed from /our_curators.

Thank you!

PR #1938, opened 2022-12-01
Removing some com curators from the cohort list

ramGranell:
This is related to ticket #1922. The following community curators were removed from the list: 7678, 7730, 7745, 7747. You can see the current ones here: https://deploy-preview-1938--fairsharing.netlify.app/community_curation/our_curators


allysonlister (approved):
Thank you this is correct!

Issue #1814, opened 2022-09-01
Adding organisation relationships when importing from ROR


ramGranell:
As we are importing fields such as alternative names and ROR ID without this appearing in the IMPORTATION ROR form (#1803), we can also import the relationships that exist in the ROR DB. In ROR, there are three types of relationships between organisations: Parent, Child and Related. We can import the first two.
Then, when an organisation is created, we can check the existing ROR parent and child relationships, e.g. Org_1 is parent of Org2 with ROR_id2. We need to check if there is an organisation with ROR_id2 in our DB. If it exists and it does not break any rule (duplicated relationship or loop over parent/child relationships), the relationship can be created when the Org_1 is saved.
This ticket may require changes in the API as well.

ramGranell:
As we want to only create the relationship if the parent or child of the new organisation exists, we need to look for it. The way to search is using the ROR id as we are not going to perform at this moment a search based on name or URL.

PR #1952, opened 2022-12-12
Rg adding ROR parent child relationhips for organisations#1814

ramGranell:
This is related to this ticket #1814 and basically it is that when importing a ROR organisation that has any relationship (parent/child) with an organisation whose ROR is included in our database, then the relationship is imported. 

To check it you need to create a new organisation, importing from ROR .For example you can import any of the child organisations that appear here https://ror.org/052gg0110 or https://ror.org/042nb2s44 (note that you cannot create an organisation that it is already in our database), then after creating it you can visit the parent organisation and you will see the relationship automatically imported from ROR. In the same way, you can try importing an organisation and later importing its parent, e.g. First I create https://ror.org/047d2d387 and later you create one of the parents that do not exist e.g. https://ror.org/01v29qb04 You can explore any example form ROR that you like.


ramGranell:
Sorry, but this feature is not working until we have the server also updated with the code. Let's wait until 13/12/2022 to update the server as it is currently used with fairassist queries.

ramGranell:
This is working now as I updated the server code and the data is properly transfered

allysonlister (approved):
This works as expected, thank you!

knirirr (approved):

Issue #1927, opened 2022-11-28
Form and updated communities page to show tools that are making use of FAIRsharing


allysonlister:
# General request
ROR has a lovely [page](https://ror.readme.io/docs/ror-integrations) that showcases those resources that are using their API or using the FAIRsharing metadata in some other way (e.g. including our IDs/links in their resource). 

We've been thinking for some time of a way in which we can also showcase those who integrate our metadata with their projects, and would like to implement something similar to this. Without getting into any technical detail, this is how we would like it to work:

1. User visits our [communities](https://fairsharing.org/communities) page and sees the list of existing tools under "Adopters" and wants to showcase their tool and how it uses the FAIRsharing API.
2. User finds a link on the page that says "Let us know how you're using FAIRsharing" (or some variant thereof), clicks on it (see below for text)
3. This takes them to a google form that they provide the details of their usage of FAIRsharing
4. The technical team are notified when the form is submitted, and they check it for completeness and that it is in scope
5. "magic" happens and the page on our static site is updated with the new card. 

Note that we've recently implemented cards for our community curation page (https://fairsharing.org/community_curation/our_curators) and that perhaps some of this code could be repurposed for this task?

# Placement
The Adopters section (that is reached via its Adopters "tile") in https://fairsharing.org/communities should now be structured as follows:

A broad range of [stakeholders](https://fairsharing.org/stakeholders) come to FAIRsharing from across all research domains. However, adopters use FAIRsharing specifically to do one or more of the following:

 i Educate their users/community on the variety of existing standards, repositories and policies that exist in FAIRsharing and actively encourage them to submit/claim records, where relevant;
 ii Recommend resources by registering their data policy, and then link it to standards and/or databases [recommended](https://fairsharing.org/search?page=1&isRecommended=true) in the policy;
 iii Create a [Collection ](https://fairsharing.org/collections) that pulls together a list of standards and/or databases around a given domain of interest relevant to them. 

Adopters are generally representatives of institutions, libraries, journal publishers, infrastructure programmes, societies and other organisations or projects that in turn serve and guide individual researchers or other stakeholders on research data management matters. Adopters display a FAIRsharing logo on their websites with a link from their website to our homepage. 

In general, our adopters fall into the following broad categories. If you don't see your project on the list of adopters, please get in touch.

Funders (section link that jumps to what is currently the first table in this section)
Publishers (section link that jumps to what is currently the second table)
Organisations (section link that jumps to what is currently the third table)
[...]

PR #1984, opened 2023-01-06
Communities 1927

knirirr:
This is not yet ready for review; once https://github.com/FAIRsharing/fairsharing.github.io/issues/1927#issuecomment-1372130362 has been resolved we can proceed with modifying this branch accordingly.


allysonlister:
Thanks for this! While I'm aware this is not ready for review yet, it is looking good and I will start contacting a few selected tools that make use of FAIRsharing so we can launch this with some real content.

In the meantime, please would you make the following updates to the text I've asked you to add in https://deploy-preview-1984--fairsharing.netlify.app/communities ?

1. I made a mistake in my [original comment](https://github.com/FAIRsharing/fairsharing.github.io/issues/1927#issue-1466602186) where I said the "Organisations" link should go to the first table - it should, in fact, be an internal link to the third table. I've updated the comment accordingly, and please could you update the static page?
2. Could the section that currently looks like this:
image
 instead be a series of buttons that stretch across a single line of the page, to make them stand out?
3. Could each of these new subsections have a subsection title in the same way "Adopters" is, but in a smaller size?
Thanks!

allysonlister:
Oh, and also, the new tiles for the tools look great, but it doesn't seem as though they have all of the information yet. There is a placeholder for a description, but not for the other fields (method of use, type of resource, status of integration). I'm happy for you to explore options about how to show this, but perhaps some sort of tags in the same way as we have for the Early Adopters in the community curators?

allysonlister:
Thank you! some comments:

1. I've asked @proccaserra to provide a better image for FAIRcookbook - I wonder if he accidentally uploaded the wrong one.
2. We need a way to also show the tool in use if any of our form submitters choose to upload such a screenshot. Perhaps if they click on the tile? I can provide such a test screenshot for FAIR cookbook below, which shows a link through to FAIRsharing
3. The homepage for the tool is missing - perhaps hyperlink the name of the tool?
4. For some reason there is no email associated with the FAIR cookboook tile?

Otherwise looking good! Thanks!

image

knirirr:
(1) Thanks.
(2) I'm afraid I don't have any ideas for that at present. 
(3) I've added a link to the tool name (a typo had caused the link to go missing). 
(4) None was provided in the contact information.

allysonlister:
Thanks!

In relation to item 2, perhaps you can ask Prakhat or Dom for advice?
In relation to 4, I agree this is an issue - because we tried to consolidate name, contact email and organisation to have fewer form items, it leaves us open to this kind of issue. However, we said we'd use all fields publicly unless otherwise specified, so this information:

image

Can be used. In this particular case, let's presume @proccaserra missed out this info accidentally, and please use fairplus-cookbook@elixir-europe.org which is their official contact email.

Thanks!

allysonlister:
Also, there seems to be some kind of colour filter on the logos, making them look darker than they are originally. Can that be looked into please?

knirirr:
> Also, there seems to be some kind of colour filter on the logos, making them look darker than they are originally. Can that be looked into please?

Indeed. Here it is:

https://github.com/FAIRsharing/fairsharing.github.io/blob/communities_1927/src/views/Static/Community/Community.vue#L305

Without it, the text is very hard to read.

allysonlister:
That makes sense. Is it possible to apply a vertical gradient of some kind as well, so that as you go "up" to the top of the tile, the dark bit disappears? Sorry that I am not wording it correctly...

allysonlister:
This is looking good thanks - looks like you still need to add the cookbook email (see above) but otherwise I think this is good for now. If you ever have any other ideas about the gradient, please feel free to implement them!

allysonlister:
image_2023_02_09T22_25_18_771Z
Here is the image Dom provided for FAIR cookbook. Once that is done, and the cookbook email (see above) has been added, I'm happy to approve this - thanks!

allysonlister (approved):
Looks great! We can add it and then encourage more people to reply to the form :) Thanks!

Issue #1985, opened 2023-01-09
Update Governance


allysonlister:
At https://fairsharing.org/communities#governance, please can you:

1. remove Molly Cranston and Guillaume Wright from our Advisory Board
2. *copy* (not move) Rebecca Grant and her information from the RDA Chairs section to the Advisory Board section, alphabetical by first name. She's replacing Molly and Guillaume.

Thanks!

PR #1990, opened 2023-01-11
Requested changes for #1985.

knirirr:
If there ends up being no Netlify preview then names are displayed in the same order they're defined in the JSON data file, thereby allowing review.


allysonlister (approved):
the preview worked, thanks! All looks good.

Issue #1989, opened 2023-01-10
Button in the curation dashboard to provide data sorted out by creation date


ramGranell:
The data should be provided by the API [#663 ](https://github.com/FAIRsharing/FAIRsharing-API/issues/663) There can be the data provided by https://github.com/FAIRsharing/FAIRsharing-API/issues/660 . It also needs to download the total edits as agreed in https://github.com/FAIRsharing/FAIRsharing-API/issues/659

PR #1992, opened 2023-01-13
modifying curation panel to download more files

ramGranell:
This is related to this ticket https://github.com/FAIRsharing/fairsharing.github.io/issues/1989 to download the files with count of records created by month and records edited by month. To test see the download buttons that are in the bottom of the curator panel:
1. RECORDS CREATED BY MONTH 
2. RECORD EDITS BY MONTH


knirirr (approved):
I'd have been tempted to move "system messages" above "records without DOIs", but this all works as expected.

ramGranell:
> I'd have been tempted to move "system messages" above "records without DOIs", but this all works as expected.

We can modify the order if curators think it is better to change it.

allysonlister (approved):
looks good to me, thanks!

PR #2017, opened 2023-02-16
Sigma network graphs

knirirr:
This now renders network graphs with Sigma.js, for an enormous speed increase on the client (at the expense of the server). 

Please note:  This work is not yet ready for requests for changes from users!


knirirr:
Outstanding issues are still:

- [x] Buggy behaviour on page reload or when back button pressed.
- [x] Filtering by distance fails after a few button presses. 
- [ ] Images on nodes don't show. 
- [x] Filtering by status must be added. 
- [ ] Switches are needed to filter by type...  (really? If so, perhaps not urgently?)
- [x] ...and by registry. 
- [ ] SVG graph downloading is still missing. 
- [x] Rendering start/stop switch should perhaps be removed (stops other switches from working) (now hidden except from curators).

allysonlister:
I'm re-writing the unfinished items from above here, together with a priority setting from my subjective viewpoint; I understand there are many other factors, e.g. complexity of the task. You'll notice some additional items too.

Outstanding issues are still:

- [x] Buggy behaviour on page reload or when back button pressed.
- [ ] Images on nodes don't show. _high priority_
- [x] Filtering by status must be added. _high priority_
- [ ] Switches are needed to filter by type... (really? If so, perhaps not urgently?) _medium priority_
- [ ] SVG graph downloading is still missing. _medium priority_
- [x] coloured relationships. See below for priorities when >1 relationship between two nodes; although I'm equally happy if you want to just display the last relationship loaded. - _low priority from a time perspective, but it is a feature I'd really like to have_
- [ ] Search in graph (e.g. node name) - please tell me if this should be a separate ticket. _low priority_

**Coloured relationships:** Please can we have the coloured relationships back. If you prefer to implement a "display only the most recent relationship" method when >1 relation exists between two nodes, that's great. If you'd rather have an order of priority, then here they are, though I think this may be difficult to implement. When you have >1 relationship between two nodes, choose the one to display from the list below that is highest on the list (==lowest number).

1. deprecates
2. recommends
3. collects
4. profiles
5. accepts
6. outputs
7. shares code with
8. shares data with
9. implements
10. extends
11. related_to

knirirr:
My plan was to proceed with merging and closing this PR once we have a new server, which I'd expect to take a few weeks.
Anything unfinished by then could go into one or more tickets at that time.

knirirr:
Comments from a meeting:

```
4 registry colours: #4678A3 (databases) #A046A3 (policies) #A37146 (collections) #29AE9E (standards)
status and hops both in FAIRsharing mid blue / pale grey when deselected
And use abbreviation instead of full name
```

PR #2025, opened 2023-02-23
Changed display of additional info. items.

### LINKED ISSUE #2021: Cross References not displaying their note text
**Describe the bug**
Cross references are displaying hyperlinked URLs, but where a note is present, the note string should be hyperlinked using the URL value

**To Reproduce**

https://doi.org/10.25504/FAIRsharing.pdwqcr
image
image

You can see there are notes, but they're not being displayed

Thanks!

knirirr:
The component modified here is one that's meant to process data only based on definitions from schema files, and so it's not something that would do customised displays for specific fields. For that, those fields would need to be suppressed from "additional info" and displayed in a custom component elsewhere.

Whilst looking into this I discovered that this component was in fact not doing what it should, and was only displaying one out of the various fields available.

If you are looking at the output rather than the code itself then please look at a variety of records to see if anything unexpected turns up in the additional info. section of record pages.


allysonlister (approved):
I am indeed looking at the behaviour, and all seems well. I've checked a few records with a few different cross reference fields. Thanks!

ramGranell (approved):
I've also checked some records and they seem fine to me.

Issue #2027, opened 2023-02-24
New tools


knirirr:
This ticket is to add his response to the tools questionnaire into the relevant JSON for the site.

@allysonlister we don't have an organisation entry for https://fairdata.systems/home/ (Mark Wilkinson)- how would you like to handle that? Presumably it could be created and assigned to Mark, but I am not sure to which records it should be linked and how.

allysonlister:
This is a really good point, actually. In theory, organisations don't have to be added to a record, just to a user. Let me add it to his user profile at https://fairsharing.org/users/866 (as this is the profile with the same email as listed in the form), and then you can use that one.

Ugh, or I would add the organisation to the user's profile, but I can't because orgs now require countries, and the "New Org" form that curators have access to from a user's profile doesn't have a country field.

I will cheat by adding the org to my account, then removing it, then adding it to Mark's.... Please could you see about creating a ticket to add a country field to that "new org" popup that curators see? As only internal curators see it, and we don't use it often, this can be a low priority ticket.

The organisation is now present in the system as https://fairsharing.org/organisations/3899

thanks!

knirirr:
Other responses have since come in and so these could be added as part of this ticket.

knirirr:
I'll close this in favour of https://github.com/FAIRsharing/fairsharing.github.io/issues/1927, as the question about Mark's organisation has been resolved.

Issue #2030, opened 2023-02-27
URL for the Tool tiles


allysonlister:
When I click on the new Tools button the URL it takes me to is [https://fairsharing.org/communities#](https://fairsharing.org/communities#) which seems odd; although it is correctly taking me to the Tools tiles section, the URL it displays does not have anything after the hashtag. This may just be my browser (Firefox on Ubuntu, Chrome on Ubuntu)

I note that https://fairsharing.org/communities#tools works properly, if I manually type it...

PR #2032, opened 2023-02-28
Adding new tools #2027

ramGranell:
Here are the new tools information:
https://deploy-preview-2032--fairsharing.netlify.app/communities#tools


knirirr (changes requested):
I suggest adjusting height at https://github.com/FAIRsharing/fairsharing.github.io/blob/master/src/views/Static/Community/Community.vue#L326 until the text for the DSW is accommodated.

allysonlister:
image


Thank you! Some notes:

1. Charles has asked that we also add Harvard Medical School to the list of organisations for the bio.registry. Please can you do that?
2. there is some wrapping/overspill on the DSW tile
3. RDMkit and FIPwizard's logos look a bit odd, as you can't see most of them. Perhaps if they don't look good when put into the tile, you can try adding some padding around the image in order to show it better? Alternatively, check the form responses as often they say that they can provide alternative icons if those don't work; or, email the tool devs and ask for a different shape of icon. In the meantime we can progress with these icons, but this is something you can check before asking for review, in general :)
4. Both DSW (I think?) and FIP wizard have provided screenshots of the tool in action. is this work part of this ticket, or will be done separately?
5. the URL that is shown when I click on the Tools button is still https://deploy-preview-2032--fairsharing.netlify.app/communities# and not communities#tools
6. We need a way to communicate the new anchors for each tile to our 3rd-party devs. Perhaps an addition to each tile that just says "link to this tile" or similar? Is that part of this ticket, or not?

Thanks!

knirirr:
Point no. 4 will have to be a separate ticket at a later date. 
Point no. 5 shouldn't be addressed here - I need to find time to look at it. 
Point no. 6 will need a separate ticket.

allysonlister:
Thanks - after 1-3 are dealt with please let me know and I'll review again.

ramGranell:
> Thanks - after 1-3 are dealt with please let me know and I'll review again.

1-2 are performed and emails were sent (FAIRsharing account) asking for square logos for 3.

knirirr:
It looks like that DSW box needs a bit more height still.

allysonlister:
I agree with @knirirr that a bit more space might be needed for the DSW tile.

and separately, before closing this ticket, please make sure appropriate new tickets are made (as necessary) for the other items, thanks. https://github.com/FAIRsharing/fairsharing.github.io/issues/2030 already exists for item 6, the other two may already be parts of existing tickets?

knirirr (approved):

knirirr:
> the other two may already be parts of existing tickets?

Yes, #1927 and #2030. 

New PRs can be opened when those are addressed; this PR can be closed after merging but not any tickets.

allysonlister (approved):
thank you!

PR #2033, opened 2023-02-28
Closes #2030.

### LINKED ISSUE #2030: URL for the Tool tiles

knirirr:
This adds the tools tag requested in #2030, and some other tags, and also fixes an un-noticed bug (wrong link for organisations).


allysonlister (approved):
thank you - looks good!

PR #2039, opened 2023-03-03
Rg square logos

ramGranell:
I only modified two of the logos (FIP Wizard and RDMkit) of the tools https://deploy-preview-2039--fairsharing.netlify.app/communities#tools


allysonlister (approved):
looking good, thanks!

PR #2057, opened 2023-04-24
JSON was in fact a string in Python example.

knirirr:
A user reported that they got a 401 from using the Python example code. I think the reason is that the payload line has somehow been formatted as a string instead of JSON.


ramGranell (approved):
Here, there is only a change in the documentation. It would be hard to see the error if we cannot access the user Python script.

knirirr:
> Here, there is only a change in the documentation. It would be hard to see the error if we cannot access the user Python script.



Indeed. If you want to try it, you could copy and paste the example from the docs. Here's the entire new version:

```

import requests
import json

url = "https://api.fairsharing.org/users/sign_in"

payload={"user": {"login":"your_username","password":"your_password"} }
headers = {
  'Accept': 'application/json',
  'Content-Type': 'application/json'
}

response = requests.request("POST", url, headers=headers, data=json.dumps(payload))

# Get the JWT from the response.text to use in the next part.
data = response.json()
jwt = data['jwt']

url = "https://api.fairsharing.org/fairsharing_records/1"

headers = {
  'Accept': 'application/json',
  'Content-Type': 'application/json',
  'Authorization': "Bearer {0}".format(jwt),
}

response = requests.request("GET", url, headers=headers)

print(response.text)

```

Issue #2059, opened 2023-05-04
Updating our Stakeholder Board


allysonlister:
Please could you add Richard Brown to our stakeholder board list (alphabetically by first name) as:

Richard Brown, BBSRC UKRI

I will add a comment when I have his ORCID as well.

allysonlister:
Please note that Richard does not have an ORCID, so we can proceed as-is.

PR #2061, opened 2023-05-05
Update communityPageData.json

ramGranell:
Adding new stakeholder, #2059 You can see the new stakeholder advisor here: https://deploy-preview-2061--fairsharing.netlify.app/communities#governance 
I needed to include an exception when there is not an orcid.


prakhyatox (approved):

allysonlister (approved):
Thank you!

Issue #2058, opened 2023-05-03
Searching organisation by ROR id


ramGranell:
When searching (filtering) organisation (https://fairsharing.org/organisations/) , being able to use the ROR id field (it is not a mandatory field).

knirirr:
I'll close this as it appears to have been merged.

PR #2062, opened 2023-05-09
Rg organisation search ror #2058

ramGranell:
This is for ticket #2058 . To check it go to the organisations list (https://deploy-preview-2062--fairsharing.netlify.app/organisations) and see that the ROR field is included (not all the organisations have a ROR link).

Note that before developing in the production, last version of the server needs to downloaded and a reindex is needed (I think that "Organisations.reindex" is just needed).


allysonlister (approved):
looks great!

knirirr (approved):

Issue #2060, opened 2023-05-05
Graphs: hop-specific filtering


allysonlister:
Please note: the FAIRsharing Assistant is the current priority, therefore this has been marked as low priority and doesn't need to be done until the initial version of the Assistant is complete.

The new relationship graphs are fantastic! I've had quite a few people comment on them, and everyone seems very happy.

I would like to have a change in how the filtering works, please: I would like to be able to filter by registry type, independently, at each "hop" level.

This would allow, for example, a collection graph to show just the collected databases in the "one hop", then all standards in the "two hop", and all databases in the "three hop". The result would be that we would see only those standards that are directly linked to the collected databases, and then be able to explore **all** databases that are directly linked to those standards.

As another example, let's say we're starting with GenBank, and I'd like to find out all of the databases that are linked to any standard that is directly linked to GenBank. I'd set the "one hop" to standards, then the "two hop" to databases. Does that make sense?

Thanks!

allysonlister:
Please note that this ticket is still active and development should continue. It was noted, however, that this is not a trivial (or even necessarily possible) task for the current graph library that is in use, and therefore a solution may take some time.

PR #2064, opened 2023-05-16
Add loads of extra toggle buttons for nodes. #2060

knirirr:
This implements the functionality requested in #2060, but the button layout for registries is broken on small screens.


knirirr:
N.B. the current layout represents the limits of what I can do in terms of styling.

allysonlister (commented):
Thank you so much for this! I'm using https://deploy-preview-2064--fairsharing.netlify.app/graph/2521 as an example here...
image

The issue here is that all standards in the "2 hop" are still being shown. In this case, what we'd like to see are only those standards linked to whatever has been selected in the "1 hop". Does that make sense? Thanks!

image

knirirr:
Ah, I see. Unfortunately I can't do that. Would you prefer to keep this branch as it is, or discard it and use the graphs as they currently are?

allysonlister:
This is a key feature for our graphs that our users really would like.

Is there any way to traverse a graph that you're displaying in a "live" manner? You are already hiding/showing nodes depending on certain characteristics. Is it possible, even with additional coding, to have a rule that nodes are only shown if they are connected to an already visible node?

knirirr:
We can't generate the graph on the fly and there doesn't appear to be any means of handling this request within the rendering engine, though I have already been investigating it (not quite finished yet). Please think on what you would prefer in the event that this can't be done, no matter how much it is wanted.

allysonlister:
Thank you for investigating, I do hope you can find a solution.

A lengthier option might be pre-computing each of the different options the user might generate (that is a LOT of combinations, I know!) at the same time as you pre-compute each graph already.

knirirr:
Generating multiple graphs will not be feasible, I'm afraid. 
Please let me know your preference for (1) keeping this branch as it is, or (2) discarding it and sticking with the graphs as currently active on the site.

allysonlister:
Let's talk about this in the next dev meeting before making any decisions. Thanks!

allysonlister:
If it cannot be done with this library or codebase, then we cannot use this branch, thanks. We'll still have a chat about longer-term options on Monday, though, as this is a key feature.

Thanks!

Issue #2073, opened 2023-05-26
Some additional info not displayed...


knirirr:
...unless it is set to "yes".
More info. can be seen on this pull request comment:

https://github.com/FAIRsharing/fairsharing.github.io/pull/2072#issuecomment-1564071257

PR #2074, opened 2023-05-26
Removed restriction on showing non-yes values. #2073

knirirr:
This fixes non-yes values not being shown for some additional info. fields.


ramGranell (approved):
I think that all the desired values are displayed.

PR #2077, opened 2023-05-31
Added new licence component for collection page #2069.

### LINKED ISSUE #2069: Display licence information for 'collection' record
Collections only show a subset of fields, and this is deliberate. 
However, it would be good if the licences could be displayed, when the field is filled in (for example: https://fairsharing.org/3541)
It could be something like this:
licence-collection

Thanks !

knirirr:
This completes #2069.
N.B. there may be multiple licences so this displays them in a list, like countries. 

I'd be interested in any comments on the getField implementation in the new component.


prakhyatox (approved):

delphinedauga (approved):
Ok, thanks !

PR #2081, opened 2023-06-08
Hidden records 2080

knirirr:
This adds a component which will tell users to log in if a record is hidden and then, if they don't have permission to view it, to wait until the curators approve it.


allysonlister:
Sorry - I'm getting a 500 error when I try to load a record (1245) as it has a hidden record linked to it and I was going to review using that... I can't seem to see the netlify build at all...

knirirr:
Other than that a database permission had changed, I can find no issues.
Do you meant that the link to the build is not appearing for you?

allysonlister:
Apologies, and yes, the link to the build doesn't work for me.

allysonlister:
Or rather, it's working now... So, I try https://deploy-preview-2081--fairsharing.netlify.app/search?q=dublin%20core and I get 3887 records - that's the majority of the database. Is the search working properly in this build?

knirirr:
I don't know about searching on this system - I'd have to investigate, but this is not a high priority as it's not a production system. 
The client-side feature I have added is unrelated to search so that can still be looked at.

allysonlister:
OK - as long as the search in this build would not be replicated if it were to be merged against master? As any search term I use returns what seems to be all records, which is not ideal....

I will, in the meantime, have another look at the hidden record part of the ticket on the netlify build

knirirr:
I've not changed anything related to the searchFairsharingRecords query in this PR:
https://github.com/FAIRsharing/fairsharing.github.io/pull/2081/files?file-filters%5B%5D=.js&file-filters%5B%5D=.json&file-filters%5B%5D=.vue&show-viewed-files=true

allysonlister:
Apologies, but although the build looks ok and I can search, when I try to load just a single page (https://deploy-preview-2081--fairsharing.netlify.app/1245) I get the 500 error on the build so can't check anything. Similarly, trying via its 'alternate' url also gives a 500 (https://deploy-preview-2081--fairsharing.netlify.app/FAIRsharing.3nx7t)

I can view records by clicking on them in a search page such as /databases, but this doesn't help as searching for 'dublin core' returns the entire registry rather than just the one I need.

After messing about for a while getting it to work, even https://deploy-preview-2081--fairsharing.netlify.app returns a 500 error for me, at least for a minute, and then I am allowed back in.

I can't check anything this way right now, unfortunately!

knirirr:
The best guess I can make here from what we can see here is that Netlify is misbehaving; the first load of each page gives the expected behaviour but then crashes. The 500 is not coming from our API or, apparently, our client code.

allysonlister (approved):
This is working correctly, thank you!

prakhyatox (approved):

PR #2087, opened 2023-06-16
Show community badges when they are the only "data process and condition" present.

knirirr:
This adds a missing check for the presence of the offending field.


ramGranell (approved):
Community badges appear

delphinedauga (approved):
Ok, thanks :)

PR #2096, opened 2023-07-07
Changed some text, added bold and newlines.

### LINKED ISSUE #2090: Change secondary tagline content/layout on homepage banner
Please can we update the secondary tagline (shown in light blue here):
image

Here is the text, please note the bold and the newlines.

Guides **consumers** to discover, select and use these resources with confidence
Helps **producers** to make their resources more visible, more widely adopted and cited
Powers **third party tools** by providing trustworthy content to promote standards and database

However, while we really would like the bold, we are completely open to the best way to display this text under the primary (in white) tagline, and are happy for you to suggest your favorite couple of options.

Thanks!
Allyson

knirirr:
This changes some text. There is HTML in the text which is being rendered by the client, but the bold parts don't llook very bold to me.
FWIW I'd prefer everything on one line, but I suspect no-one else will.


allysonlister:
Apologies for the delay!

Susanna and I are happy with each sentence on its own line, thank you.

As I cannot see the bold either, please could you move that portion of the ticket to a new ticket (low priority) for resolution at some point when we're not so busy? We'd like to have the bold elements visible, but it doesn't need to stop this PR from being approved and can be completed later.

Thanks!

allysonlister (approved):

PR #2099, opened 2023-07-19
Update jumbotronData.json

ramGranell:
This is a second brand to try to sort out this: #2090 
The text is changed and the bold words appears in the subtitle


allysonlister (approved):
thank you!

Issue #2109, opened 2023-07-26
ROR logo in records


allysonlister:
In the same way as we have the small orcid logo linked to a user's profile, please could we have a small ROR logo next to the name of any organisation in a record that has a ROR link? When clicking on the logo, it will take you to that organisations' ROR page. Thanks!

ramGranell:
An example of ROR link to organisation can be found here https://catalogue.cioospacific.ca/dataset/ca-cioos_16ae186b-9d99-42cf-b18d-09f9bb0501d7 (in contacts section). I copied a screenshot
![exampleROR](https://github.com/FAIRsharing/fairsharing.github.io/assets/39826987/f9ea3d66-9016-4085-a446-5b698192d172)

The problem that I see is that we have the link to our organisation page over the name and now a new link over the ROR icon is added (for these organisations with existing ROR link). This may be not very intuitive for people that is not familiar with our system and ROR . In the previous example both the link over the name and the ROR icon goes to the ROR website, however in the user maintainer we have both links as well (our user profile and ORCID).

allysonlister:
Thanks! I see your point. However, I'm happy (if you are) to carry on as we have for our usernames; the link on the organisation babe being to our organisation late, and the ror logo going to the ror page. Thanks!

ramGranell:
After checking the response by the API server I saw that the field ror_link is not sent in the graphql response with all the information about the record. In the field organisationLinks.organisations there is an array of organisations in which each organisation contains the following fields:  "name", "id", "homepage", "types", "urlForLogo" and "countries" but not "rorLink". I think that a server ticket is needed to obtain that field.

PR #2130, opened 2023-09-08
Linkable ROR icon for organisations in the record page

ramGranell:
This is related to ticket #2109. A good example it is this record https://deploy-preview-2130--fairsharing.netlify.app/FAIRsharing.wkggtx Check the organisations in that section.

@knirirr , you can note, that Codacy complains saying that "Current Status: Not Analyzed " meaning that it is not able to analyse the code.


allysonlister:
Thank you - the links look good, and I like the size and placement of the ROR logos.

When you click on the ROR logo it loads by default in the current page, while when you click on the FS organisation link it loads by default in a new tab. I am agnostic about the default behaviour, but I would like both of these clicks to behave the same. If you would like an opinion (though I am happy to hear other ideas), I'd suggest opening both FS orgs and ROR links in a new tab.

Thanks!

ramGranell:
> Thank you - the links look good, and I like the size and placement of the ROR logos.
> 
> When you click on the ROR logo it loads by default in the current page, while when you click on the FS organisation link it loads by default in a new tab. I am agnostic about the default behaviour, but I would like both of these clicks to behave the same. If you would like an opinion (though I am happy to hear other ideas), I'd suggest opening both FS orgs and ROR links in a new tab.
> 
> Thanks!

This is changed and a new tab is open when pressing the ROR icon.

allysonlister (approved):
All looks great, thanks!

knirirr (approved):
In this case I think that the codacy warning can be ignored (I had a long discussion with their support about this sort of thing some time ago).

Issue #2135, opened 2023-09-18
Tags: None -> N/A


knirirr:
On a record page there will be "none" if there are no tags of a particular type found, e.g domains. This is to be changed to "N/A" for those tags only.

PR #2136, opened 2023-09-19
Changed "None" to "N/A" to help confused users. #2135

knirirr:
This deals with #2135 by changing the label if no tags are shown (as discussed elsewhere).


allysonlister (approved):
looks great!

Issue #1903, opened 2022-11-03
Add "gained skills" to our community curator tiles


allysonlister:
Additional Feature for https://fairsharing.org/community_curation/our_curators

At some point in the coming weeks, we are going to want to tag our curators with some additional skills/expertise tags from [terms4FAIRskills](https://obofoundry.org/ontology/t4fs)

These tags will be visible when the user clicks on the blue "+" icon in the top right of a curator's icon on the page. Each tag should be in the style of the "Early Adopter" tag that already exists, just visible from the blue overlay once the "+" is clicked. Below is a mockup of what I'm looking for please:

![image](https://user-images.githubusercontent.com/143586/199695080-a5138098-0510-4b04-96af-b54006a0e7da.png)

They will have the following fields (an example fictional tag is included):
Tag name|tag URL|date achieved
|---|---|---|
|using persistent identifiers|http://purl.obolibrary.org/obo/T4FS_0000492|20.10.2022

n.b. the "date achieved" will not be visible on the curator profile tiles, but we want to store it internally in case we want to refer back to it in future.

knirirr:
What is the source of these data to be? Specifically, are we expected to get this directly from another site via an API or will the data be added to our users' accounts manually, e.g. by adding the fields to https://github.com/FAIRsharing/fairsharing.github.io/blob/master/src/data/communityCurationCohorts.json ?

If it's just the latter then it shouldn't be too difficult to include a field in each entry such as:

```
{
  ...
  "year_active": ["2022"],
  "gained_skills": [
    {
      "name": "using persistent identifiers",
      "date": "20.10.2022",
      "url": "http://purl.obolibrary.org/obo/T4FS_0000492"
    }
  ]   
},
```

Displaying it is another matter, of course...

allysonlister:
I was only intending to add them directly to the json as described above, thanks!

prakhyatox:
By when data for curators are expected?

allysonlister:
As soon as the curators complete the requirements for gaining these tags - so it depends on how quickly they curate ;-)

allysonlister:
Our first curator has met the challenge I set for getting the first tag. I realise that @prakhyatox is busy with other work, but perhaps someone else could take a look at this and complete it if possible?

Thanks!

allysonlister:
Hi all,
 
[...]

PR #2142, opened 2023-09-25
Gained skills for Champions #1903

ramGranell:
This is the first attempt for ticket #1903 . If you go https://deploy-preview-2142--fairsharing.netlify.app/community_champions/our_champions
There are two champions with gained skills: Kyle Copas (one skill) and Annie Kristensen IMPORTANT, this is fictional data and I need to remove fictional data from the file src/data/communityCurationCohorts.json until the real data is available.

I also included ticket #2143 request here. There are two cases when the champions is tagged as curation expert:
1. It has the Curation Expert but not Early Adapter (see champion Gabiel Pelletier)
2. It has the Curation Expert  AND Early Adapter  (see champion Joe Miller)


allysonlister (approved):
This looks as I would expect. When putting into production, please could you link this value to Lindsey's tile. Thank you!

knirirr:
@allysonlister am I correct in thinking that you want this to be deployed without any gained skills data being available, and for these to be added later?

If so then this deployment can be done once @ramGranell has had a moment to remove the fake demo data.

knirirr (approved):
This code seems to do the trick, but shouldn't be merged until the fake data have been removed, of course.

allysonlister:
This should be deployed with NOTHING added to anyone yet, other than the Curation expert for Lindsey. Thanks!

ramGranell:
> This should be deployed with NOTHING added to anyone yet, other than the Curation expert for Lindsey. Thanks!

All data about Champions as it was but only adding Curation Expert foir Lindsey.

Issue #2150, opened 2023-10-02
Please change "curators" to "champions" when attributing within records


allysonlister:
![image](https://github.com/FAIRsharing/fairsharing.github.io/assets/143586/b9df02b0-bef6-443c-a85a-4470c200bef9)

Please change the word "curators" to "champions". Thanks!

PR #2151, opened 2023-10-03
Ticket #2150 Update CommunityCuratorInfo.vue

ramGranell:
I left the word "community", then it is "community champions" is it ok?


allysonlister (approved):
There's no netlify build but, checking the change in the code itself, this is exactly correct, so happy to approve. Thanks!

ramGranell:
> There's no netlify build but, checking the change in the code itself, this is exactly correct, so happy to approve. Thanks!

Sometimes when the change is very small Netlify does not work properly. Anyway, the only change I made is the word as I commented. You can see the change in the code in the label "Files changed" (ignore the documentation files and go for the last one).

Issue #2152, opened 2023-10-03
Not appearing new field in the edit system


ramGranell:
**Describe the bug**
The new regular expression field that is created for Standard-Identifiers is not appearing. I check it and it giving back from the server

**To Reproduce**
1. Go to 'https://fairsharing.org/317/edit'
2. Click on 'Additional information'
3. There is not the new field

**Expected behavior**
The field should be appear and be editable

PR #2153, opened 2023-10-03
Dealing with showing additional information that it is a String

ramGranell:
This is related to record #2152 
To check you can edit any Identifier and see the result in the page
as an example this record: https://deploy-preview-2153--fairsharing.netlify.app/317
As I'm taking the special case of being the additionalInformation a string, the rest of cases should work correctly.


src/components/Records/Record/AdditionalInfo.vue
```diff
@@ -88,9 +88,15 @@ export default {
       // if received node is not an array
       if (!isArray(selectedNode)) {
         this.tempData[key] = []
-        Object.keys(selectedNode).forEach(item_key => {
-          this.tempData[key].push({[item_key]: selectedNode[item_key]})
-        })
+        // Special case this is a string
+        if (typeof selectedNode === 'string' ){
+            this.tempData[key].push({"": selectedNode})
```
knirirr:
```suggestion
            this.tempData[key].push({"Value": selectedNode})
```

...on the grounds that it looks a bit weird to me with an empty key.

tests/unit/components/Records/Record/AdditionalInfo.spec.js
```diff
@@ -88,6 +89,11 @@ describe("AdditionalInfo.vue", function(){
                 {
                     "a": []
                 }
+            ],
+            "justString": [
+                {
+                    "": "Hello"
```
knirirr:
```suggestion
                    "Value": "Hello"
```

knirirr (approved):

Issue #2163, opened 2023-10-12
Remove extraneous full stops from related records


knirirr:
As shown in the image below, there appear to be some full stops which are being rendered between linked record cards. These ought to be removed.

![Screenshot from 2023-10-12 11-23-05](https://github.com/FAIRsharing/fairsharing.github.io/assets/3583375/1a2882ea-29c7-4414-9d08-4ec8bb9230ad)

PR #2167, opened 2023-10-18
Removing point in record related items #2163

ramGranell:
This is related to #2163. It was a point that someone left. I spent some time investigating the vue component...


knirirr (approved):
Truly an outstanding PR!

Issue #1964, opened 2022-12-14
Persistence policy


knirirr:
A static page with our persistence policy is required. A link should be added to the footer (which may be re-arranged later).

knirirr:
This has been deployed.

PR #2183, opened 2023-12-01
Persistence policy 1964

knirirr:
This should address #1964, together with https://docs.google.com/document/d/1w8W_43FvhcRB7gROCvTDPpxH71OCOHJvDS-dkt-0xsg/edit#heading=h.a3l6nn4600ec

I propose that once the document is finished (there's still a link to add to it) that it is uploaded to Zenodo and we link to it directly. I have not put the text of it on the policy page so we don't have to keep the page in sync with the definitive document. 

Any thoughts on this, @allysonlister ?


allysonlister:
This all sounds good, thanks. I believe we have a FAIRsharing community/collection on Zenodo that we can use for this purpose.

prakhyatox (approved):

allysonlister (approved):
Looks great, thanks!

PR #2185, opened 2023-12-05
Remove twitter 2184

knirirr:
I've concealed the Twitter link. Just in case the issue is ever resolved I'll leave the necessary code in the server for now.


allysonlister (approved):
The appropriate bit seems commented out, so thank you! I can't confirm with a netlify build, but I don't think I need it in this instance. Thanks!

knirirr:
Thanks. I checked it locally as well and the Twitter button isn't visible.

Issue #2213, opened 2024-01-18
Hide deprecated records by default in graphs


knirirr:
Presumably they are still to be included in the data, but the default option for showing them should be toggled off.

prakhyatox:
The Deprecated button is off but graph still shows the deprecated records.
Needs further investigation on it to check whether Sigma can handle such kind of customisation.

The issue is kept open until ```Advanced Search``` https://github.com/FAIRsharing/fairsharing.github.io/issues/2139 is completed.

PR: https://github.com/FAIRsharing/fairsharing.github.io/pull/2233

PR #2214, opened 2024-01-19
Turn off deprecated graphs by default #2213.

knirirr:
This only deals with the UI, specifically #2213.
Anything relating to whether nodes are correctly shown as deprecated or not is in https://github.com/FAIRsharing/FAIRsharing-API/issues/861, which has been deployed on the dev server and is awaiting a quieter time to go into production.


allysonlister:
Thanks. When I go to e.g. https://deploy-preview-2214--fairsharing.netlify.app/graph/5119, the deprecated records are shown by default, and there is no toggle to have them on or off. Am I looking in the wrong place?

knirirr:
That's the correct place, but not the correct result when you go there. Netlify is currently broken for me and I get something different.
This PR definitely works locally and for some people via Netlify, and does what it's supposed to, so I'll merge it anyway I think.

PR #2215, opened 2024-01-23
Org page redesign 2159

knirirr:
https://deploy-preview-2215--fairsharing.netlify.app/organisations/3112 is a suitable organisation to show how this change looks.

N.B. logos can only be added to records by:
1. A record edit form (not yet done and out of scope for this PR).
2. A tedious process of adding by hand or script. 

Any requests for a different style preference (e.g. location or [size](https://youtu.be/5AxwaszFbDw?feature=shared&t=18) of logo) will probably have to be addressed by @prakhyatox.


allysonlister:
Thank you! I've added @delphinedauga as her thoughts would be most welcome.

Please could we:

- [x] add associated countries
- [x] have the logo appear in a similar place, size and style as our 'regular' records
- [ ] have the justification/formatting for the various field names and values match our 'regular' records

https://github.com/FAIRsharing/fairsharing.github.io/issues/2216

I think that's everything for now - I like how you've made a white background for the general metadata rather than the cream of the existing records so you can tell at a glance that it's different from those records.

knirirr:
@allysonlister please would you put the last of those comments in a separate ticket; it will require work on both server and client which will require some thought, so I can't do it as part of this PR. 

The logo is where it is due to limitations on my skills. The justification is as it is because I think it good and appropriate. Changing them will require assistance from @prakhyatox. 

I should be able to deal with the countries.

allysonlister:
Thank you! Happy to have the logo item moved to a separate low-priority ticket for @prakhyatox to deal with when he has a moment, if that's ok?

I've moved the other item to another ticket.

allysonlister:
It's nice to see blue pills for clickable links, and grey pills for non-clickable ones. Ultimately would be great to get countries and types clickable, and I know you're working on that.

After the justification is sorted, I'm happy to approve, thank you.

(would be nice see what @delphinedauga , who's in a different time zone, has a chance to reply)

src/components/Organisations/TagChips.vue
prakhyatox, line 1: Moving this file to shared folder (```src/components/Records/Shared```) would be better option just like ```KeywordTooltip.vue```  in the same folder, since these both files are being used in advanced search results too

src/utils/stringUtils.js
```diff
@@ -38,6 +38,8 @@ export const cleanString = {
     }
 };
 
+// Removed from Organisation.vue, but I've not deleted it in case it is needed again.
```
prakhyatox, line 41: No need to comment this function, since it is removed from the Organisation.vue, it won't impact anything.

tests/unit/components/Records/Record/CollectionRecord/searchCollection.spec.js
```diff
@@ -88,7 +88,7 @@ Record.state.currentRecord["fairsharingRecord"] = {
     registry:'Collection'
 };
 
-describe("SearchCollection.vue", function(){
+describe("SearchOrganisationRecords.vue", function(){
```
prakhyatox: Was this change required? Seems the file original file name has not been changed.

src/components/Organisations/SearchOrganisationRecords.vue
prakhyatox, line 1: Could not see its test cases

src/components/Organisations/TagChips.vue
prakhyatox: Could not see its test cases for this file

src/components/Organisations/SearchOrganisationRecords.vue
prakhyatox, line 1: Could not see its test cases

prakhyatox (changes requested):

src/utils/stringUtils.js
knirirr: Are you suggesting that the function simply be deleted?

tests/unit/components/Records/Record/CollectionRecord/searchCollection.spec.js
knirirr: This appears to be a mistake in copying the file for the new component.

src/components/Organisations/SearchOrganisationRecords.vue
knirirr: This relates to the mistake mentioned in the previous comment.

src/utils/stringUtils.js
prakhyatox: No, it should remain like it is. Not to comment that function.

knirirr:
I have moved the logo positioning to https://github.com/FAIRsharing/fairsharing.github.io/issues/2217 - this can safely be done later as logos aren't available at the moment anyway.

@allysonlister I've been thinking about what you mean by justification, and I think you might be referring to the gap between the label and text as seen in the screenshot below. Is that correct?

Screenshot from 2024-01-23 17-35-10

allysonlister:
Thanks @knirirr that's exactly what I was referring to - apologies for being unclear!

prakhyatox (approved):

allysonlister (approved):
Looking great, thank you!

delphinedauga:
Very nice page ! Easy to understand.
Some comments : 
- organisation logo a little bit bigger
- ROR logo also a little bit bigger, I miss this information at the first look
- can we see somewhere the relation with the related records ?

allysonlister:
Thanks Delphine!

> * organisation logo a little bit bigger

Organisation logo work has moved to another ticket :-)

> * ROR logo also a little bit bigger, I miss this information at the first look

That would be nice actually, I agree!

> * can we see somewhere the relation with the related records ?

The relation is within the results panel in the bottom half of the org page - if it's too small we can make it bigger?

knirirr:
ROR icon: The one we use elsewhere is 24px and the largest they have is 32px. It will look blurred if enlarged. Even repacing the 24 with the 32 doesn't look great on the page, but I have done it anyway.
To make the relation more obvious I have placed an outlined v-chip around it rather than increase the font size.
This should, I hope, be enough to finally merge this PR...

allysonlister:
thank you, this is good for me. @delphinedauga if this is good for you, please could you approve the PR?

Issue #2211, opened 2024-01-11
Governance page changes


allysonlister:
Please remove the following board members from the governance page who are now alumni. Note that this needs to be done by Monday 16 January:

1. Imma Subirats
2. Amye Kenall
3. Luiz Olavo Bonino
4. Dominic Fripp

Thank you!

allysonlister:
Here's an associated update too:

1. Put  Nick Everitt and Matthew Cannon on separate lines in alphabetical order, both affiliated with Taylor and Francis
2. Please can you also update "Rebecca Grant, F1000" to instead be "Rebecca Grant, Taylor and Francis"

PR #2224, opened 2024-01-26
Removed some board members, edited others. #2211

knirirr:
Changes have been made as per #2211.

If Netlify doesn't build anything then the changes can be seen here instead: https://github.com/FAIRsharing/fairsharing.github.io/pull/2224/files


allysonlister (approved):
looks good, thanks!

PR #2231, opened 2024-02-07
Pub doi fix 2230

### LINKED ISSUE #2230: What type of zenodo DOIs are accepted by our Record Edit Interface?
I have just had to manually add https://zenodo.org/doi/10.5281/zenodo.4705073, as the "import from doi / zenodo" function did not succeed.

To reproduce, I've given another similar DOI as this one is now in our system

image

Try to add https://zenodo.org/doi/10.5281/zenodo.10490288 via the Import from doi/zenodo button and you'll get the above error message "DOI not found!"

Not sure if this is intended, e.g. if the type of publication associated with this DOI is not one that we allow, or if it is a bug...

thanks!

knirirr:
This adds a tooltip and attempts to handle parsing returned JSON from Zenodo without failing. 

#2230


ramGranell (approved):
It works for both
doi/10.5281/zenodo.4705073
and
https://zenodo.org/doi/10.5281/zenodo.4705073

Issue #2228, opened 2024-01-29
Download button not working as expected in collections


allysonlister:
When I use the download button at https://fairsharing.org/4566?page=1 I get a file (attached) of everything in FAIRsharing and not just the 51 records in this collection.
[fairsharing_download_1706537488.csv](https://github.com/FAIRsharing/fairsharing.github.io/files/14085752/fairsharing_download_1706537488.csv)

knirirr:
I'm not sure how to proceed with this yet but I can confirm that the client isn't sending search terms to the server, but only on Collection pages. It's probably because the search components were shoe-horned into the record page after the original design.

knirirr:
Done and will be deployed on the next merge to dev.

PR #2232, opened 2024-02-08
Hacked around download button search query issue #2228.

knirirr:
If you're trying this out, please check that you also find that:
- Downloading collection results is correct.
- Downloading the results of a proper search is still correct.


prakhyatox (approved):

allysonlister (approved):
looks good thank you!

PR #2233, opened 2024-02-08
changing Deprecated active flag to false

prakhyatox:
Ticket: https://github.com/FAIRsharing/fairsharing.github.io/issues/2213

Previous PR:https://github.com/FAIRsharing/fairsharing.github.io/pull/2214

Although, the ticket was closed by implementing changes in above PR but it somehow got reverted and the changes were reflecting in dev.

### DIFF
--- a/src/data/networkGraph.json
+++ b/src/data/networkGraph.json
@@ -86,7 +86,7 @@
     {
       "name": "Deprecated",
       "color": "#27aae1",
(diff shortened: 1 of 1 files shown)


knirirr:
I don't think this can be merged at present; changing this field doesn't prevent the deprecated nodes from being rendered intitially (which is unexpected).

PR #2235, opened 2024-02-09
Publications in collections 2218

### LINKED ISSUE #2218: Make publications visible in collections if present
https://fairsharing.org/5027 has some publications relevant to the collection. At the moment, we allow adding of publications to collections, but have never displayed them as there hasn't been a need.

Please can you investigate a simple way to display publications in collections, but only if they are present? Note that as with other record types, the publications may have one that is the publication to use for citation.  We don't need an empty or "N/A" field if there are no publications, as the majority of collections won't have them.

Happy for you to reuse existing record code, but placement may be an issue, so happy to see other ideas too.

Thanks!

knirirr:
This displays publications on collection records in the same style as other records, but higher up the page so it is not concealed by the scrollable record list.


allysonlister (approved):
Looks good, thank you!

PR #2248, opened 2024-02-16
File uploader 2182

### LINKED ISSUE #2182: File uploader
A report has been received that images added to a record are not being uploaded, with the uploader code claiming that they are over the maximum size when that is not the case.

Not all users are experiencing this. For those with difficulties, some report it working as expected in Firefox but not Chrome, and others the opposite.

knirirr:
It should now be possible to add logos to fairsharing_records and also delete them, reliably, on all browsers. Thus, #2182 should be resolved by this PR. 

Please check both whether the code itself looks OK, and also whether the component behaves as expected.


allysonlister:
I've tried on Firefox and Chromium, all fine.
When I use more-than-normally wide logos, it overlaps a bit with other fields, but not necessary to fix this unless it's easy to do so.
image

I did also notice that it's not necessary to use the delete logo button; one can upload a new logo successfully without deleting the old one. However, this isn't a problem at all.

Functionality is great, thanks.

allysonlister (approved):

knirirr:
Thanks. Re. the deletion of logos, I thought there might be a circumstance that someone would want to remove one without having a replacement handy. 
@prakhyatox any thoughts on limiting width in that logo preview field?

prakhyatox (approved):

delphinedauga (approved):
Tried on Opera, Safari and Chrome, no pb, everything works well

PR #2250, opened 2024-02-19
Added least recently updated record https://github.com/FAIRsharing/FA…

knirirr:
…IRsharing-API/issues/900

This accompanies the server code in https://github.com/FAIRsharing/FAIRsharing-API/pull/905.


ramGranell (commented):
Two things:
1. In the netlify, when accessing to the curator panel. Now, there are few seconds (5-6s) that the curator panel is not showing but a generic FAIRsharing screen. There is not any sign or anything loading and I thought it was not loading.
2. Also, the link of the record least recently updated is https://deploy-preview-2250--fairsharing.netlify.app/function()%20%7B%20%20%20%20[native%20code]%7D326 that does not go to a correct record.

knirirr:
This is the best I can manage given the constraints of the existing component.
I'll open another ticket for refactoring but this will probably have to be put off for now.

ramGranell (commented):
Hi, now the two things that I commented are sorted out. It goes directly to the panel that load the page and the link is working. The only thing is that the record suggested does not exit longer, 
https://deploy-preview-2250--fairsharing.netlify.app/326
Do you think it can be a deprecated or deleted record? Do you think it makes sense for the server to check only ready records?

knirirr:
That wasn't originally specified, but seems like a sensible change to the request.
However, it can't be done here, as it will all be server-side. So, I'll merge this one and change that separately on the server.

Issue #2254, opened 2024-02-22
Displaying collections recommended by policies


knirirr:
Following on from https://github.com/FAIRsharing/FAIRsharing-API/issues/909, changes will be needed to ensure that collections recommended by policies are displayed on record pages. It's not entirely clear how this will work given the structure of that part of the codebase.

PR #2255, opened 2024-02-22
Show recommended collections on a policy page. #2254.

knirirr:
The only way I can see to work around this change is to add a "related collections" tab only when the record being displayed is a policy. This will handle recommended collections.


prakhyatox (approved):

allysonlister:
https://deploy-preview-2255--fairsharing.netlify.app/3491
image
So does this mean that any collection linked to a policy via the 'collects' relationship (e.g. a collection collects a policy) will be present twice, as shown in this screenshot?

knirirr:
> So does this mean that any collection linked to a policy via the 'collects' relationship (e.g. a collection collects a policy) will be present twice, as shown in this screenshot?

It does indeed! This part of the code is not very good and is hard to both read and modify. Changing this is out of scope for this PR, though.
As I'm guessing you don't like seeing the collection in two places I've added a hack to remove it (and will have to ask @prakhyatox to review again, apologies).

prakhyatox (approved):

allysonlister (approved):
All the behaviour I've tested looks good, thank you.

Issue #2249, opened 2024-02-19
Add counts to summary boxes on front page


allysonlister:
![image](https://github.com/FAIRsharing/fairsharing.github.io/assets/143586/4bd210cd-c7c5-4076-b41d-e0b742f0a35a)
Please could we have a count of institutional policies here too? We only have 9 at the moment, but that number should be growing in the next year and I'd like to make it more obvious that we cater to that group.

Thanks!

allysonlister:
I've changed the subject as I've seen a few more similar updates to perform please... to summarise all:

1. add institutional policy count to the policy box
2. add journal publisher policy count to policy box
3. add count of institutional repositories (https://fairsharing.org/search?userDefinedTags=institutional%2520repository)  to the summary box

thanks!

knirirr:
>     3. add count of institutional repositories (https://fairsharing.org/search?userDefinedTags=institutional%2520repository)  to the summary box

I'm not sure which box this is - would you mind confirming when you have a moment, please?

allysonlister:
At the bottom of here:
![Screenshot_20240223-111238.png](https://github.com/FAIRsharing/fairsharing.github.io/assets/143586/127abc8c-1b01-4a85-89a3-5f412f2da17b)

I realise it's not its own registry but it's a useful count for our users. Thanks!

knirirr:
This will appear on the main site next time a merge to master takes place (which will be when https://github.com/FAIRsharing/fairsharing.github.io/pull/2255 is complete).

PR #2256, opened 2024-02-23
Added new counts to home page. #2249.

knirirr:
The new fields are added according to #2249, but the three record count boxes are no longer of the same height.

Edit: Now fixed.


allysonlister (approved):
looks great thanks!

PR #2262, opened 2024-02-29
Clicks on the coloured bits 'work' now #2261.

### LINKED ISSUE #2261: Adding tags to a record - UI
Some users don't understand that to add tags to a record they must click "add new term(s)". It's not clear why, but I think they're clicking on the coloured bars showing tags, finding nothing happens, and giving up.
A couple of possibilities spring to mind.

1. Change the button to say "edit tags" (which might be clearer for them).
2. Add a tooltip to the component directing them to that button.

knirirr:
Now, clicking on the coloured area (which I what I think the users who give up and email for help are doing) will expand the tag editing interface just as the button does. 

This is a cunning idea I had in relation to #2261


delphinedauga (approved):
Too good! I approve!

allysonlister (approved):
fantastic, thank you!

Issue #1240, opened 2021-09-28
Editing Organisations via the UI


allysonlister:
While the ability to edit organisations exists, it has not yet been implemented in the UI. When there is time to add this feature, having the ability to edit organisations records such as https://beta.fairsharing.org/organisations/1961 would be great.

Thanks very much!

allysonlister:
Please note that this temporarily on hold while we have further discussions about organisations as a curation object. I will comment again when we're ready to move on this. Thanks!

knirirr:
Before we proceed with this, we ought to determine who can edit organisations. Presumably, given that anyone can add themselves to one, it should be curators only?

allysonlister:
yes please, let's do curators and higher - thanks! That way Champions can also edit them.

PR #2264, opened 2024-03-07
Organisation editing

knirirr:
This PR offers the following:

- Logo upload when users create an organisation.
- Logo displayed on organisation pages. (*)
- Curators can edit organisations from their pages. (**)

(*) As best I can; someone else will have to change it if not to reviewers' tastes. 
(**) Except users and records, which are done separately, and parent/child organisations, which, if required, I would rather save for a separate ticket later. 

#1240


src/components/Editor/Organisations/LinkOverlay.vue
```diff
@@ -528,17 +533,29 @@
...
+              if (_module.organisation.urlForLogo) {
+                _module.menus.newOrganisation.delete('logoData');
+              }
+              else {
+                _module.menus.newOrganisation.logoData = {}
+              }
+              return;
+            }
+            _module.logoLoading = true;
+            let convertedFile = await toBase64(logo);
```
prakhyatox, line 553: Conversion toBase64 could be called on click of ```Save new organisation``` button.

src/components/Editor/Organisations/LinkOverlay.vue
```diff
@@ -528,17 +533,29 @@
...
-            let reader = new FileReader();
-            reader.addEventListener('load', async (event) => {
-              this.menus.newOrganisation.logoData = {
-                filename: this.menus.newOrganisation.data.logo.name,
-                data: event.target['result'],
-                content_type: this.menus.newOrganisation.data.logo.type
-              };
-            });
-            reader.readAsDataURL(this.menus.newOrganisation.data.logo);
+        'menus.newOrganisation.data.logo': {
```
prakhyatox, line 537: This watch could be removed.

src/lib/Client/RESTClient.js
```diff
@@ -605,6 +605,25 @@ class RESTClient {
         return response.data;
     }
 
+    /**
+     * Create a given organisation
```
prakhyatox: Should be ```Edit a given organisation```, since the function is about edit organisation

src/views/Organisations/Organisation.vue
```diff
@@ -183,21 +196,224 @@
         </v-fade-transition>
       </v-col>
     </div>
+    <!-- Edit existing organisation -->
```
prakhyatox, line 199: If possible can Edit organisation be a separate component, giving the complexity and size of the file.

src/components/Editor/Organisations/LinkOverlay.vue
knirirr: On trying to remove this I see why it was originally done; if I attempt to access `this.menus.newOrganisation.data` within `createNewOrganisation` then it's always equal to `{}` (not so if it's done in the watcher).

src/views/Organisations/Organisation.vue
knirirr: Having had  a look I don't think that this could easily be done.

src/components/Editor/Organisations/LinkOverlay.vue
knirirr: I've addressed this one here: https://github.com/FAIRsharing/fairsharing.github.io/pull/2264#discussion_r1521762363

allysonlister:
Looking great. What are the conditions on the logo upload? I tried a jpg (rather large tbf) and it just stayed red with no message. A png was successful.

Here was the random image I unsuccessfully tried. It is completely unsuitable in size, which was one of the reasons I tried it. Just wondering about the reason it was rejected, and if we need to have some small text about the requirements for successful upload near or within the Logo upload text?
RDA 10th anniversary Official Zoom Background

knirirr:
The size limit is 1048576 bytes, i.e. ~1 MB.

src/views/Organisations/Organisation.vue
prakhyatox: Created ticket for the task: https://github.com/FAIRsharing/fairsharing.github.io/issues/2267

prakhyatox (approved):

allysonlister:
> The size limit is 1048576 bytes, i.e. ~1 MB.

Thanks - is it worth putting this in the text somewhere? As it's only curators, it's not mandatory, but might be helpful to add any constraints to the text for forgetful people like me.

delphinedauga:
I've found this bug when I try to edit: the scrollbar on the side doesn't move the window up or down. I find this bug in Chrome and Opera. However, it works in Safari

https://github.com/FAIRsharing/fairsharing.github.io/assets/59688426/ead9b070-8f05-4f26-b467-6fa87d63a8fd

allysonlister:
FYI I did not have this bug in Firefox on Ubuntu.

knirirr:
I have also been unable to reproduce this.

allysonlister:
Actually, having just tried the scrolling again I'm getting the same behaviour on Firefox as @delphinedauga mentioned - I think I just didn't have my window small enough last time. If i make my window small enough that I would need to scroll, then no scrolling happens even though the little bar on the right moves up and down.

The message for the logo is great, thanks

knirirr:
This is not really a bug; because I used an overlay scrolling will affect the element beneath which you won't see if the overlay is filling the screen. I've now changed it to a dialog which will change what is scrolled.
Resizing the browser window to something very small may still cause issues, but I was able to use this component successfully on  an emulated iPhone (both as overlay and dialog) so I think it will be OK.


Screenshot from 2024-03-18 15-03-41

delphinedauga:
It works great! Perfect!

delphinedauga (approved):

allysonlister (approved):
works great, thanks!

PR #2297, opened 2024-04-18
Deletion of organisations via UI. #2293.

### LINKED ISSUE #2293: Deleting organisations
Staff members are able to edit organisations, which is great! Please can we also delete organisations e.g. via a button next to the edit button on the organisation page, with a confirmation dialog too? Today https://fairsharing.org/organisations/4342 was added and this is clearly advertising, and I'd prefer to not have to trouble the tech team to remove such items.

thanks!

knirirr:
Adding a delete button (with associated functionality) to the organisation page.


src/views/Organisations/Organisation.vue
```diff
@@ -631,6 +672,17 @@ export default {
       );
       await this.getCountries();
       this.loading = false;
+    },
+    async deleteOrganisation(del) {
+      if (del) {
+        let data = await restClient.deleteOrganisation(this.organisation.id, this.user().credentials.token);
+        if (!data.error) {
+          // Deploy to organisations page.
+          const url = new URL(process.env.VUE_APP_HOSTNAME + '/organisations');
```
prakhyatox: ``` const url = new URL(process.env.VUE_APP_HOSTNAME +'/organisations');```
  ```window.location.href = url.toString();```

is redirecting to ```localhost:8080//organisations``` thus causing site to break.

Instead a simple use of router push will solve issue
~```const url = new URL(process.env.VUE_APP_HOSTNAME +'/organisations');```~
  ~```window.location.href = url.toString();```~

Solution: ```await this.$router.push('/organisations')```

knirirr: I tried the $router.push you suggested. But, it doesn't reload the data and so it's possible to delete an organisation and be directed back to the list only to find the organisation still listed there. I suspect that a user would complain sooner or later so using window.location.href solves this. 
Is it just the double slash that's causing issues? I could fix that.

prakhyatox: I have fixed the issue using ```window.location.pathname = '/organisations'```.
I have to do force push in this case. I am not sure but the I was getting ```git pull ``` before ```git push``` and the ```git pull``` was somehow not being successfull.

prakhyatox (approved):

src/views/Organisations/Organisation.vue
knirirr: Strange that it should not be successful. Please let me know if it happens again as I'd like to have  a look in case anything stands out as a potential cause.

allysonlister (approved):
looks great, works as expected, thanks!

Issue #2296, opened 2024-04-18
Please remove "FAIRsharing tips" image from Educational


allysonlister:
As we will be focusing on the other infographics first, please will you remove or hide the following image from /educational ?

![image](https://github.com/FAIRsharing/fairsharing.github.io/assets/143586/fe901fb4-3971-4d60-8f1c-51e63513b189)

Thanks!

PR #2298, opened 2024-04-19
Hide educational 2296

knirirr:
Removed tips as per #2296 and reformatted some code to remove warnings.


allysonlister:
This looks good thanks! 

While I was checking the images, I noticed that it seems that there is a space before the "Across the research disciplines..." text - is that correct? It looks as if it isn't justified the same as the next line down:

image

knirirr:
All the paragraphs are indented like that; I've not changed anything in the layout of this page other than to remove one element from an associated JSON file:

Screenshot from 2024-04-22 11-38-51

prakhyatox (approved):

allysonlister (approved):
ok, no worries, thanks for pointing that out. Thanks!

PR #2303, opened 2024-04-26
Completed #2302

### LINKED ISSUE #2302: Add our newest factsheet to the /educational page
New DOI for our policies overview factsheet: https://doi.org/10.5281/zenodo.11059819

- [x] When the user clicks on the DOI text, they actually get a link directly to the PDF, in this particular case the URL to direct the user to is https://zenodo.org/records/11059819/files/FAIRsharing%20content_%20policies%20overview%20V1.0.pdf?download=1
- [x] When the user clicks on the "copy" icon, they are directed to the actual DOI, which is https://doi.org/10.5281/zenodo.11059819
- [x] The updated image is attached, to replace the grey "coming soon" one. FAIRsharing content_ policies (1)
- [x] The shortcut link for this new factsheet is https://fairsharing.org/educational#policies, and should work similarly to the others.

Once this is done we can promote on socials. Might we be able to have this update this week please? No worries if it is next week. thanks!

knirirr:
The new factsheet has been added as described in #2302 (i.e. the same as the existing ones). 
If there's no netlify build (this is only a data change) then the "files changed" option above will be sufficient to indicate what has been done, i.e.

https://github.com/FAIRsharing/fairsharing.github.io/pull/2303/files?file-filters%5B%5D=.json&file-filters%5B%5D=.png&show-viewed-files=true

Edit: Netlify decided to build something this time:

https://deploy-preview-2303--fairsharing.netlify.app/


allysonlister (approved):
looks great!

Issue #2295, opened 2024-04-18
Hover-over help text in records


allysonlister:
A user has recently suggested the following:

> May I suggest a hover-over info be added to the headings on the page such as "Reference URL" could mean link to a page where you tell us about linked things or it could be something you want people to cite when refering to your resource. 
I think each of those heading items on the left of the page would benefit from readily available descriptions about them (either hover-over or link-out).

So, essentially they're suggesting that, just as we have "?" in the edit interface, we should have some way of providing short descriptive sentences about each of our fields in our record view. Having lots of "?" might clutter up the interface, but could we have a sentence appear if a user hovers over the text itself?

To get a first pass at the help text, you could look at the existing help text for the edit ui and adjust it appropriately for someone viewing the record. Thanks!

knirirr:
Are you expecting there to be tooltips for everything where there is one for editing?
Personally, I'd prefer the consistent appearance of a tooltip icon for both editing and viewing, or neither.

allysonlister:
Yes, that's right, tooltips for every field where there is one in editing. we already have some tooltips with "?" in the record viewing, e.g. cross references in the additional information section. So, that style, but more prevalent across the record.

Thanks!

knirirr:
Do you expect to see tooltips for the general information section, e.g. name, abbreviation, description etc.
In the edit interface they are generally present but not for status or registry (which I would have thought needed them).

allysonlister:
Yes, tooltips for the general information section, all items if possible. Once we've done that we can look at the other sections and see where it makes sense to add them

knirirr:
This PR covers the general information section: https://github.com/FAIRsharing/fairsharing.github.io/pull/2305

Further PRs will be necessary at a later date (once Vue3 is dealt with) to complete tooltips for other record sections.

PR #2305, opened 2024-04-29
Record page tooltips

knirirr:
Ticket: https://github.com/FAIRsharing/fairsharing.github.io/issues/2295


knirirr:
Here's where the tooltip definition text can be found:
https://github.com/FAIRsharing/fairsharing.github.io/blob/record_tooltips_2295/src/data/recordTooltips.json

delphinedauga (approved):

allysonlister:
Here is the current view on a screen-maximised version of dryad in both the netlify and prod builds. It seems that every field EXCEPT homepage, year, type and registry are all aligned differently from how they are at present.

Once the alignment is fixed I will approve. In the meantime I have also updated the tooltips slightly [here](https://github.com/FAIRsharing/fairsharing.github.io/blob/record_tooltips_2295/src/data/recordTooltips.json). Did I do that correctly? 

Thanks!

Netlify:
Screenshot from 2024-05-13 10-21-19


Prod:
image

allysonlister:
"maintainers" has an empty tooltip - please could you add the following text for it?

"A list of developers of this resource who have claimed ownership of the record both for attribution and to perform updates the record."

knirirr:
Your modification was successful; the application runs locally.
I have added text for maintainers.
I'm aware of the alignment but it's not within my capabilities to fix, unfortunately.

prakhyatox (approved):

ramGranell (commented):
There are two things:
1. Content of the tooltips:
I think that the tooltip of Maintainers is missing "to", i.e.  "...perform updates to the record.". 
Maybe in the tooltip of "Taxonomic Range" it would be helpful to indicate that this field is only for "Life science related records", i.e. this is not apply to other disciplines instead the name is generic as taxonomies exist in other areas. I suppose that this field is heritage from Biosharing. Writing a thing like "This is only applied to records related to life science."?
2. When I go the icons of support, organisations, data processes it appears a blue circle with a question mark and expect to have a tooltip but nothing appears. Anyway, this may not be related to this ticket and tooltips.

allysonlister:
I agree with Ramon on his first point.

With regards to his second point, I do not see any tooltips or grey question mark icons outside of the General Information and Cross References. However, note that in the originating ticket I said we can start with the General Information and move to the other sections after that. Given the time constraints for getting started on vue3, can I suggest that a new ticket is spawned from https://github.com/FAIRsharing/fairsharing.github.io/issues/2295#issuecomment-2085344914 that can be given a low, post-vue3 priority please?

Finally, some elements of the General Information tab are still missing "?" below is an example from https://deploy-preview-2305--fairsharing.netlify.app/3495 which shows that Contact, Organisations and Reference URL are missing "?":
Screenshot from 2024-08-29 13-52-49

Please can you carefully go through all possible fields and make sure they have them? The main issue with the screenshot above is that Reference URL is just for collections (and so may easily have been missed), and Contact and Organisation are ONLY added to the General Information when it is a Collection, because Collections don't have the subsections where these elements would normally go.

Thanks!

knirirr:
I've added the missing sections and corrected the text.
Re. the ticket, I'd think it best to keep the existing one open and to open another PR when there's time to address other parts of the application.

allysonlister (approved):
thank you - the update looks good, and happy to progress with the rest of the ticket as you suggest

delphinedauga (approved):

ramGranell (approved):

Issue #2316, opened 2024-05-09
Slow profile load


knirirr:
Please see https://github.com/FAIRsharing/FAIRsharing-API/issues/961 for details.
This will require harmonising the means by which user profiles are loaded so that the same API call is used at all times.

PR #2317, opened 2024-05-09
Provides one solution to #2316.

knirirr:
This may not be the best solution. I need to look at query pagination before proceeding further.


knirirr:
This branch fixes the slow loading problem. 
There is a more elegant solution of making the edit_events call paginated on the server and calls to it linked to the buttons on the v-data-table, but I'm not currently sure how to do that.
So, I propose using this branch to work around the problem for now and creating a ticket for pagination that I can look at when (if) I get time.

allysonlister (approved):
I have tested logging in with user/pass and not via orcid as that doesn't seem to work - may just be because I'm on a dev server. I've also tested the load button. This works much better now, it's about 10 seconds to load my user profile (and of course the approx 30 secs to load all of my edit events).

knirirr:
Indeed, logging in with ORCID isn't possible on netlify builds - apologies.

prakhyatox (approved):

PR #2332, opened 2024-05-22
User password reset

knirirr:
The only solution I could come up with was to add a message telling users that they should log out and request a password reset if they see a request for their current password but don't know it. There's also a button to click to take them to the right place.

Users who don't read this and email us instead can thus be told by anyone to click that button or otherwise log out before they can perform the reset. 

The following was successful for me:

1. Log in (with ORCID if on a local machine, not available for Netlify).
2. Go to profile, select password change.
3. Click  on the red "log out and reset" button.
4. Type email into box.
5. Click on link in email.
6. Specify new password.
7. Log in.
8. ...
9. Profit!


ramGranell (approved):
Locally, I was not able to completely try it as maybe is not sending the email or i don't receive it to the mailinator.com account to the orcid.sandox. The email is displayed in the screen.

knirirr:
> Locally, I was not able to completely try it as maybe is not sending the email or i don't receive it to the mailinator.com account to the orcid.sandbox.

You have to look in the server console on your system to see the email - it will not be sent when the system is running in development mode.

ramGranell:
Yes, the email appears there with an url that it goes to a json file with 

message | "Please supply a new password and password confirmation"
reset_password_token | "FvgunEe9fsf1EqJ-cSXz"

knirirr:
> Yes, the email appears there with an url that it goes to a json file with
> 
> message | "Please supply a new password and password confirmation" reset_password_token | "FvgunEe9fsf1EqJ-cSXz"

Replace `localhost:3000` with `localhost:8080` in the URL.

ramGranell:
I forgot to reply but all seems to work locally when following previous steps.

allysonlister (approved):
looks good to me!

PR #2341, opened 2024-06-07
Rg new champions#2339

ramGranell:
https://deploy-preview-2341--fairsharing.netlify.app/community_champions/our_champions
Note that Claudia picture does not have too much resolution and that in the dev server her ID count and organisation does not exist but they exists in production.


allysonlister:
I had missed out Hugh's LinkedIn profile in the intakes spreadsheet, but there were already linkedin profiles for both Hugh and Clara, and mastodon account for Hugh that haven't been included in their champion tile. Please add these items and then should be good.

ramGranell:
> I had missed out Hugh's LinkedIn profile in the intakes spreadsheet, but there were already linkedin profiles for both Hugh and Clara, and mastodon account for Hugh that haven't been included in their champion tile. Please add these items and then should be good.

I have added them now and works correctly

PR #2368, opened 2024-07-17
Reverting change in MaintenanceRequests.vue

ramGranell:
This is related to ticket FAIRsharing/FAIRsharing-API#1034


knirirr (approved):

ramGranell:
After testing it, I wrote some processing notes in the ownsership request in the curation panel but they were not saved when I reload the page.

knirirr:
> After testing it, I wrote some processing notes in the ownsership request in the curation panel but they were not saved when I reload the page.

That's why I originally removed that option.
The problem will lie on the server but I wasn't able to determine what it was.

PR #2405, opened 2024-10-01
Used a different method to check that the search string isn't empty.

### LINKED ISSUE #2314: Extraneous "required" text
The search box on the main page of fairsharing.org is sometimes showing a red "required" string even when it isn't being used. This will, no doubt, be a side-effect of the implementation of advanced search. 

Screenshot from 2024-05-08 14-24-08

knirirr:
In order to get the "required" text to not show when it shouldn't I've replaced form validation with a dialog telling users to provide a search term if they try a general search without having done so. Of course, it won't be triggered if they click the advanced search button without providing text. 

See #2314 for details.


allysonlister:
I cannot see the netlify build, but I have just one comment.

Would it be easier to just have the user who clicks on "Search" just taken to https://fairsharing.org/search which just presents the user with a list of all content? I often do this action thinking it will work, and it doesn't... :)

knirirr:
> Would it be easier to just have the user who clicks on "Search" just taken to https://fairsharing.org/search which just presents the user with a list of all content? 

I'm not sure. Is that what you would prefer?

allysonlister:
If we're changing the behaviour such that the odd 'Required' does not appear as often, AND it is sensible to have the behaviour I mentioned above solve this issue, then yes, that would be great. Not sure what the underlying issue is that's causing this, of course. Generally speaking, I'd like to be able to click on 'Search' and get all results if the text box is empty. thanks!

knirirr:
> odd 'Required' does not appear as often

It won't appear at all now. Previously, clicking on the search box but not typing anything and then doing anything else (e.g. moving the cursor away) would cause "required" to show. 

> AND it is sensible to have the behaviour I mentioned above solve this issue, 

That is the question, indeed.
Because the result is paginated it's not as slow as it could be:
https://fairsharing.org/search?fairsharingRegistry=Standard,Database,Policy,Collection&searchAnd=false&page=1
Is this a useful feature, though?

allysonlister:
For me, yes, I find myself manually typing /search all the time into the top bar. That's a sample size of one, admittedly.

knirirr:
I have changed this PR to load all records when the search button is clicked.

prakhyatox (approved):

allysonlister:
Although the behaviour when clicking 'search' is correct, when I type something in and hit 'return' then it displays the appropriate search for a moment, and is then superseded by the complete registry search. In firefox. Here is a recording:
[Screencast from 2024-10-08 10-20-54.webm](https://github.com/user-attachments/assets/c1b9b0fb-29bd-4f29-9959-231e77a0ee52)

allysonlister (approved):
looks good thanks

Issue #2402, opened 2024-09-30
Adding F-UJI to our tools page


allysonlister:
Line 9 of our spreadsheet of form responses includes The F-UJI team's information on how they use FAIRsharing. Please could this be added to our tools page? Thanks!
[Zulip comment with spreadsheet link](https://zulip.fairsharing.org/#narrow/dm/10,12-dm/near/57660)

knirirr:
@allysonlister is there an organisation within our database which is associated with F-UJI?

allysonlister:
Thanks - please use this: https://fairsharing.org/organisations/427

knirirr:
Thanks, I've added it.

PR #2406, opened 2024-10-01
Added tool as per #2402.

knirirr:
This adds F-UJI, though I don't know if it should be associated with an organisation in our database.


allysonlister:
I've linked the organisation (https://fairsharing.org/organisations/427) in the other comment; I'll review again once you've said you've added that, thanks!

allysonlister (approved):
looks good thanks!

PR #2409, opened 2024-10-02
Removed 'uncertain' from data curation. #2408

### LINKED ISSUE #2408: Advanced Search: data curation and checking that fields name match
The [data curation](https://fairsharing.gitbook.io/fairsharing/additional-information/database-conditions#data-curation) field in contains the following values: manual, automated, manual/automated, none, not found:
image

However, the same field in the advanced search has a different set of values:
image

In particular, "uncertain" is not the same as "not found" - they mean two different things.

Please can you fix the data curation pulldown in the advanced search, and ensure that the possible values for all pulldowns, including data curation, matches what the values are as shown in the edit UI / documentation?

knirirr:
This small change has been made (removing the extraneous 'uncertain') and, as expected, I didn't find any similar examples.

If no Netlify, then the details are here: https://github.com/FAIRsharing/fairsharing.github.io/pull/2409/files?file-filters%5B%5D=.vue&show-viewed-files=true


allysonlister (approved):
Thanks, this looks great.

PR #2419, opened 2024-10-18
Graph button hidden if no graph. #2377

### LINKED ISSUE #2377: Hide graph button
Currently, users can click on "view network graph" even if there isn't one to view.
It might be worthwhile hiding the button if there's no graph.

knirirr:
This PR hides the graph button and disables the menu entry if there is no graph to see, according to #2377. 
To prove this, compare most records (e.g. 1547) with 123, from which I have deleted the graph.

The PR for the server-side change is here: https://github.com/FAIRsharing/FAIRsharing-API/pull/1083


prakhyatox (approved):

allysonlister (changes requested):
I like the greyed-out "View relation graph" in the actions menu.

While this is functioning as I would expect, I am also thinking ahead to the report generation, where we are likely to have a button that is greyed out when the functionality is unavailable (e.g. "Log in to download report").

Bearing this in mind, would it be better to have the behaviour when there is no graph as follows?

1. for the action menu: continue to grey it out as written here, but also change the text to "No graph available"
2. for the button in the Gen Info card: change colour to pale grey and change text to "No graph available". 
3. Add a "?" next to the button in Gen Info that is there whatever the state of the graph that says: "Generate relation graphs are available for all records unless there are no relations or the graph has not yet been generated."

What do you think? We would need the greyed-out non-clickable button in item 2 so that the "?" isn't hovering on its own when there isn't a graph.

Happy to hear contrary views here.

knirirr:
I have no opinion either way, and so would be happy to make those changes (assuming I can...).

allysonlister:
Thanks - please would you go ahead with my suggestions then.

allysonlister (approved):
thank you looks great!

Issue #2427, opened 2024-11-18
Fix Twitter link (target="_blank")


knirirr:
We received a report that the link to twitter.com in our footer (presumably) uses `target="_blank"` without `rel="noreferer,noopener"`, which needs addressing.

PR #2428, opened 2024-11-18
Changed links in footer. #2427

knirirr:
This PR is for changing  `target="_blank"` links as per #2427.


knirirr:
Remaining examples can be left as per these links from @prakhyatox 

https://stackoverflow.com/questions/50709625/link-with-target-blank-and-rel-noopener-noreferrer-still-vulnerable
https://chromestatus.com/feature/6140064063029248
https://mathiasbynens.github.io/rel-noopener/

PR #2430, opened 2024-11-21
Doi not found fix 2429

### LINKED ISSUE #2429: Importing OSF DOIs via the Publications edit tab
I cannot retrieve 10.17605/OSF.IO/9r62x from the Publications page. It's an OSF DOI. When I press Import DOI it just spins and hangs.

@knirirr asked me to make a ticket, and says: Behind the scenes a CORS problem prevents [crosscite.org](http://crosscite.org/) from redirecting to [osf.io](http://osf.io/). So, not strictly a hang, but the Javascript doesn't exit cleanly and reload the buttons.

knirirr:
This should now respond properly (i.e. a warning rather than a crash) if a DOI for the wrong type of object is used.


src/components/Editor/EditPublications.vue
```diff
@@ -523,51 +523,61 @@ export default {
         } else {
           let dataPublication;
           if (Array.isArray(data)) {
+            console.log("Array!");
```
prakhyatox: Please remove

src/components/Editor/EditPublications.vue
```diff
@@ -523,51 +523,61 @@ export default {
...
             dataPublication = data[0];
           } else {
+            console.log("Object!");
```
prakhyatox: Please remove

prakhyatox (changes requested):

Issue #2444, opened 2024-12-19
Update list of Board members


allysonlister:
Mike Huerta is retiring, so please can we remove him from https://fairsharing.org/communities#governance ?

Thanks!

Issue #2442, opened 2024-12-16
Alumni Champions 


allysonlister:
For early January:

The following Champions have had their access returned to 'user' and been removed from Zulip. Please could you update their status to Alumni and make corresponding changes to https://fairsharing.org/community_champions/our_champions. 

Thank you to these Champions for all you have done for us so far!

Alex Fraser	https://fairsharing.org/users/10490
Gabriel Pelletier	https://fairsharing.org/users/7670
Kelly Cobey	https://fairsharing.org/users/9389
Priyadarshini Thirunavukkarasu 	https://fairsharing.org/users/10145
Richard Acton 	https://fairsharing.org/users/9832
Stephen Serjeant	https://fairsharing.org/users/7735

Issue #2440, opened 2024-12-11
2025: New Champions 


allysonlister:
The following Champions are new and should start in January:

### Mark
![markPic - Mark MCKERRACHER](https://github.com/user-attachments/assets/51b1bf70-5973-42a0-978b-f8dd6714d50e)
Scope: Archaeology and RDM
ORCID	https://orcid.org/0000-0002-0299-6687
FAIRsharing Account:	https://fairsharing.org/users/12756

### Jason
![Jason_Partridge_photo - jason partridge@bodleian ox ac uk](https://github.com/user-attachments/assets/192a4d46-be12-4ab3-936d-edfb9110cda4)
Scope: Institutional repositories
ORCID	https://orcid.org/0000-0001-9819-9991
FAIRsharing Account	https://fairsharing.org/users/10061

Additional information for both is available in our 'Intakes' spreadsheet - ask if you don't have the link

I will create a ticket next week about those Champions who are becoming Alumni.

allysonlister:
There are two additional new champions who just need to create their account with us and then I'll add them to this ticket as well.

allysonlister:
An additional new champion: Baptiste Cecconi:
Scope: Astronomy, planetary sciences and heliophysics
FS account: https://fairsharing.org/users/10988
ORCID: https://orcid.org/0000-0001-7915-5571
<img width="321" alt="BCecconi-picture - Baptiste Cecconi" src="https://github.com/user-attachments/assets/84b4eb88-7f7d-4f4a-bb0d-38e5721bba3c" />

knirirr:
I've added everyone above. Once we have more information then I can create a new PR for those champions.

allysonlister:
Thanks!

Here's one of the outstanding Champions. Once I tell you that we've got all of the ones I know of, is that when we'd do the ORCID attribution? (Or has that already been done?)

### Alexandra Rooney
https://fairsharing.org/users/13063
Scope: Clinical Trials & Research Data Management
ORCID: https://orcid.org/0000-0003-4975-5773
![Alex Rooney Image - Alex Rooney](https://github.com/user-attachments/assets/bf7a58f6-f0bf-4b2f-aede-0849448afa5c)

knirirr:
We've not pushed any awards to ORCID yet.
It could in theory be done at any time for any champion who has authenticated with ORCID, but it might be prefereable to wait until we have them all.

ramGranell:
[...]

PR #2445, opened 2025-01-03
Board and champions

knirirr:
This includes all of #2444, #2440 and #2442 as the changes were small.


ramGranell (changes requested):
I think there are two new Champions missing: one is here: https://github.com/FAIRsharing/fairsharing.github.io/issues/2440#issuecomment-2553433917 and the second one needs to have his FAIRsharing account created based on AL comments in the ticket.

knirirr:
> the second one needs to have his FAIRsharing account created based on AL comments in the ticket.

I'm not sure what you mean by this - which one, and which comments?
I can add the other one shortly.

ramGranell:
AL, said in the ticket about the new Champions: https://github.com/FAIRsharing/fairsharing.github.io/issues/2440#issuecomment-2545708054 
that there are new Champions that they need to create their account and she would put in the ticket. I think one of them is the one in the following comment: https://github.com/FAIRsharing/fairsharing.github.io/issues/2440#issuecomment-2553433917

knirirr:
I'm not sure I understand...
If you're suggesting waiting then there's no need - this PR can be merged if it's correct in relation to the information we actually have, and I can create another later for any new information.

ramGranell:
Well, the ticket #2440  is not completed until all the new Champions are included. Have the new Champions created their user account (we can ask about this next Monday). I think we can wait until next week to know more about this.

knirirr:
> Well, the ticket #2440 is not completed until all the new Champions are included. Have the new Champions created their user account (we can ask about this next Monday). I think we can wait until next week to know more about this.

It's not completed, but this PR is; it does everything in #2444 and #2442 as well as everything we have so far in #2440.

ramGranell (approved):

Issue #2456, opened 2025-01-15
Adding Lea as curator


ramGranell:
Adding Lea user (https://fairsharing.org/users/10801) as curator

Issue #2455, opened 2025-01-15
Add a new Board member


allysonlister:
Please could you add "Ishwar Chandramouliswaran, NIH" to the list of Board members on the website?

Thanks!

PR #2457, opened 2025-01-15
New curator

ramGranell:
This is for tickets #2455 and #2456

Check https://deploy-preview-2457--fairsharing.netlify.app/communities


ramGranell:
Lea has checked and agreed with it.

PR #2460, opened 2025-01-17
New registry 2458

knirirr:
Should allow editing relations, creating new records and viewing relations.
All that's missing is the Principles icon.


allysonlister:
Should the left-hand facets of the search results pick up the new registry?

When I run this query https://deploy-preview-2460--fairsharing.netlify.app/search?q=metric, I see that the record type includes metrics (as I would expect) record types:
image

But the registry facet does not include fairassist:
image

Thanks!

allysonlister:
The benchmark that is in the preview server is showing as a Standard, not as part of the fairassist registry:
https://deploy-preview-2460--fairsharing.netlify.app/6055

allysonlister:
Do you need me to provide a list of standards records that should be moved to the Principle record type? I see that https://deploy-preview-2460--fairsharing.netlify.app/FAIRsharing.WWI10U is still listed as a reporting guideline

knirirr:
I think that all the comments above relate to there not having been any curation of the data; I have only written the new code and created some fixtures. What I'm interested in seeing from this PR is if the creation of records and editing of relations on these new records matches what you expect. 
A list might be useful; it will involve writing a rake task to move everything. Manual editing of the data directly could be fiddly and error-prone.

allysonlister:
Ok thanks - tomorrow I'll test more of what you are suggesting .

prakhyatox (approved):

allysonlister:
* Improve principles text in the new record form
image
Text for Principles: Principles are human-readable statements that define requirements for research data management goals, and are often high-level and conceptual (e.g. FAIR and CARE).

* Benchmarks are still in Standards instead of FAIRassist in the new record form

Principles are looking good though: https://deploy-preview-2460--fairsharing.netlify.app/6156

I will continue to test

src/components/Editor/EditRelationships.vue
```diff
@@ -643,16 +656,25 @@
               search = this.search.trim();
             }
             let registries = [];
+            let types = [];
             Object.keys(this.searchFilters).forEach(filter => {
               if (this.searchFilters[filter]){
-                registries.push(filter);
+               // TODO check if this is a registry or type
```
knirirr:
```suggestion
               // Check if this is a registry or type
```

allysonlister:
image
Perhaps remove 'Text for Principles' from the above? :-)

allysonlister:
### Principles checks: record and relations

Checked via https://deploy-preview-2460--fairsharing.netlify.app/6156 https://deploy-preview-2460--fairsharing.netlify.app/6157 https://deploy-preview-2460--fairsharing.netlify.app/6158 https://deploy-preview-2460--fairsharing.netlify.app/6043 https://deploy-preview-2460--fairsharing.netlify.app/6063 https://deploy-preview-2460--fairsharing.netlify.app/6044
Principle part of exactly 0..1 Principle; Principle has part 0..n Principles

Principle part of another principle
2nd principle also part of same principle
2 levels of principle parent-child 'part of' relations

All looks good, thanks, except for the cardinality of the relationship. I should not have been able to add a second part-of relationship to a different principle at https://deploy-preview-2460--fairsharing.netlify.app/6157. However, I can't remember if you said you would be able to implement that or not (and if not, that we would have to check it via curation).
image


### Metrics

Checked via  https://deploy-preview-2460--fairsharing.netlify.app/6159 https://deploy-preview-2460--fairsharing.netlify.app/6160

measures principle: tested with https://deploy-preview-2460--fairsharing.netlify.app/6159 - the relationship can be added (and shows as a change you need to save) but it does not appear in the right-hand list of relationships, even though it is successfully added and can be seen in the record when you exit editing:
image

What's really weird with this issue of invisible 'measures principle' is that it appears in OTHER edit interfaces for metrics. After I found the above I used the same tab to open the edit interface for https://deploy-preview-2460--fairsharing.netlify.app/528/edit and as you can see below the issue is randomly in this record too. I can show you this if you struggle to recreate it, just let me know.
image

And when I went to Chromium from Firefox to try to continue the testing, the issue was propagated to yet another metric record in another browser?! https://deploy-preview-2460--fairsharing.netlify.app/597/edit
image

Could this be a edit interface issue, or could it be an issue relating to the move of the metric record type to  a new registry?


### Benchmarks

checked via https://deploy-preview-2460--fairsharing.netlify.app/6160

I'm not sure we need a reference URL here, I know that this record type is based on a collection (which has a reference URL) but I expect the benchmark homepage will fulfil this role.
image

I still need to finish the benchmark and metrics tests, but will wait until the odd relationship issue above is resolved. thanks!

knirirr:
>  Perhaps remove 'Text for Principles' from the above? :-)

Done. Took a while to find (it was only in one copy of the database).

> I should not have been able to add a second part-of relationship to a different principle at https://deploy-preview-2460--fairsharing.netlify.app/6157. However, I can't remember if you said you would be able to implement that or not (and if not, that we would have to check it via curation).

I was able to do that, there are unit tests for it and it passes. I don't currently know why that didn't happen here but my guess is that the special callback isn't being called, perhaps because it's a record_association being modified. I've made one change to the way it's called but if you still experience that problem let me know and I'll try something else.

> it does not appear in the right-hand list of relationships, even though it is successfully added and can be seen in the record when you exit editing:

The filtering (linked to the switches at the top of the linked records box) did not take type into account, only registry. I've fixed that. I'm not sure what you mean about it appearing in other interfaces but whatever it is should be fixed by the changes I just made (please tell me if not).

> I'm not sure we need a reference URL here,

I've deleted it from the FAIRassist schema and reloaded schemas on the test server.

allysonlister:
>  >    I should not have been able to add a second part-of relationship to a different principle at https://deploy-preview-2460--fairsharing.netlify.app/6157. However, I can't remember if you said you would be able to implement that or not (and if not, that we would have to check it via curation).
> 
> I was able to do that, there are unit tests for it and it passes. I don't currently know why that didn't happen here but my guess is that the special callback isn't being called, perhaps because it's a record_association being modified. I've made one change to the way it's called but if you still experience that problem let me know and I'll try something else.

I have just added two part-of relationships to again. To recreate, please

1. go to https://deploy-preview-2460--fairsharing.netlify.app/6157
2. delete both existing part-of relationships
3. add part of 'example principle' and save & exit
4. add part of 'TRUST principles' and save & exit


The other two issues I found are now working correctly, thanks.

It'll likely be next week before I progress with more checks, thanks.

knirirr:
I've managed to put that relation count constraint into the model. But, the way it will report violations of this constraint to the user is rather unhelpful and there's no way for me to correct that on the server.
It might be possible to put a second constraint in the client; I'll have to see if that is possible.

allysonlister:
1. I try to save the relationships edit tab from https://deploy-preview-2460--fairsharing.netlify.app/6157 (which has the two 'part_of' from my last test) unchanged and it says it saves successfully even though it still has 2 part_of statements.
2. I then remove the TRUST principles to test the latest updates. 
3. I try to add back in the TRUST principles, and I get this orange message, which is nice:
Screenshot From 2025-02-04 11-42-00
4. Then I hit 'cancel' and try to save and exit, as I correctly have just one part_of relationship, but instead of saving I get this red banner, despite there being nothing wrong with the record:
image

allysonlister:
After this latest oddity is resolved, I think I need to approve this and we can get it into the main body of code. Creating real content will uncover any additional issues, but I can't find any more such issues right now. Thank you!

knirirr:
I can only guess at (1), and my guess is that it is related to a combination of the following things:

- Creating a record association in the right-hand panel will create an association but not save the record. The validation of number of allowed relations occurs when this save is attempted (on the server) and also when an attempt is made to create the relation (in the client). 
- Saving the record is separate from saving the record associations.
- If a record is saved without callbacks and has two associations (with your restrictions) on it then it will save successfully.
- If a record is saved with callbacks then those associations will be validated and this validation will fail. The only message Rails deigns to pass is the one you saw. 

Executive summary: It's all a side effect of the code changing whilst you had invalid data. Starting with a record which has no relations should produce the correct behaviour.

allysonlister:
thanks - I'll take another look in a bit - there's a 500 error on the netlify build for me right now

allysonlister (approved):
I've had another look and it all looks great. I'm happy for us to continue with this now.

Issue #2452, opened 2025-01-09
Curation Panel Refactor


knirirr:
Currently, the curation panel loads everything at once and puts a lot of strain on the server.
Firstly, the query needs to be refactored: https://github.com/FAIRsharing/FAIRsharing-API/issues/1144
Then, the interface could be refactored into a tabbed one which loads each table separately as needed. 
This is probably best done before upgrading to Vue3.

PR #2474, opened 2025-01-30
Refactoring curator panel

prakhyatox:
Ticket: https://github.com/FAIRsharing/fairsharing.github.io/issues/2452


allysonlister:
This looks great, thanks! Some questions:

1. I believe I remember @knirirr saying that the "least recently updated" record at the top is a particular cpu hit
Has this been made a 'lighter' query, if necessary? Otherwise, happy to have that query 'on request' too, as the others are.
2. I think we need some plain text at the top of the first two tabs stating in a human-readable way exactly what queries are being run, as I'm a little unclear about what's happening on these two. Here are my guesses, please could you tell me if I'm right?

**Records/curator edits awaiting approval:** Why 479? That seems really high.... It appears to be all new and updated records who have had a curator or 'unknown' as their last editor. Is this correct?
**user edits awaiting approval:** 40 seems about right. It appears to be all new and updated records who have anyone other than a curator or 'unknown' as their last editor - this includes 'FAIRsharingTeam'.

I'm pretty sure I know what the other tabs are as they are unchanged, and don't require any explanatory text.

3. Is each tab only loaded upon request, to save load on the server (which I think is the reason for these changes, right)? e.g. Hidden records is only queried once I click on the tab? Or is it some other methodology? 
4. All 'Download Records' seem to run simultaneously as soon as I click on the tab. Is there some way to only run each one individually e.g. when you click a button? This will prevent all running if only one file is needed.

Thanks!

knirirr:
>     1. I believe I remember @knirirr saying that the "least recently updated" record at the top is a particular cpu hit
>        Has this been made a 'lighter' query, if necessary? Otherwise, happy to have that query 'on request' too, as the others are.

There's not been any change to the query, and it is indeed a bit of a hit.

ramGranell:
>2. I think we need some plain text at the top of the first two tabs stating in a human-readable way exactly what queries >are being run, as I'm a little unclear about what's happening on these two. Here are my guesses, please could you tell me >if I'm right?

Well, in the first tab it is indicated "curator edits awaiting approval". It means all the people that are with "curator" user role are there (curators, super-curator, developers). They are the ones that appear in the column "Curator". The number 479 seems to high as you said. I think that there are adding a lot of records that are approved but they have in the column "Curator" someone, e.g. 5236. These records shouldn't appear as they are not waiting approval.

>4. All 'Download Records' seem to run simultaneously as soon as I click on the tab. Is there some way to only run each >one >individually e.g. when you click a button? This will prevent all running if only one file is needed.

Running the calls individually was not requested in the original ticket and I think is quite complex to change to individual cals (maybe more two tabs with some specific calls). Anyway, I don't think that this is a real problem as how often curators require to enter to this tab, once per month?

allysonlister:
**Human readable text** Thanks wrt item 1 @ramGranell - I still think a brief plain text explanation would be helpful. I would also recommend changing the tab name from "Records/curator edits awaiting approval" to "Curator edits awaiting approval" as the "Records" is confusing to me - what else would it be? 

**queries**: I thought the entire point of the ticket was to reduce server load, and only load those queries that the user was asking for? Otherwise why refactor the dashboard at all? @knirirr could you provide your thoughts here please?

If we are indeed trying to make the server load lower, I suggest having as the first tab the "User edits awaiting approval" as these are definitely more of a priority than curator edits.

knirirr:
> Well, in the first tab it is indicated "curator edits awaiting approval". It means all the people that are with "curator" user role are there (curators, super-curator, developers). They are the ones that appear in the column "Curator". The number 479 seems to high as you said. I think that there are adding a lot of records that are approved but they have in the column "Curator" someone, e.g. 5236. These records shouldn't appear as they are not waiting approval.

Here's the query: 

https://github.com/FAIRsharing/FAIRsharing-API/blob/master/lib/modules/curation_info.rb#L122-L147

TD;DR: It's all records where `approved` is false and the last editor is a curator of some sort.

knirirr:
> **queries**: I thought the entire point of the ticket was to reduce server load, and only load those queries that the user was asking for? Otherwise why refactor the dashboard at all? @knirirr could you provide your thoughts here please?

That's exactly what this PR (and the associated server ones) do; previously, every single query was executed at once. If I tried looking at the curator panel on my laptop, running locally, it would crash due to being out of memory. Our development server has been affected similarly as well.

ramGranell:
> Here's the query:
> 
> https://github.com/FAIRsharing/FAIRsharing-API/blob/master/lib/modules/curation_info.rb#L122-L147
> 
> TD;DR: It's all records where `approved` is false and the last editor is a curator of some sort.

In that panel there are records that are approved, e.g. 5237 (I've just checked in the dev server and it has true in approved). I think that the second part of the query adds them ((https://github.com/FAIRsharing/FAIRsharing-API/blob/master/lib/modules/curation_info.rb#L133-L144)).

knirirr:
> In that panel there are records that are approved, e.g. 5237 (I've just checked in the dev server and it has true in approved). I think that the second part of the query adds them ((https://github.com/FAIRsharing/FAIRsharing-API/blob/master/lib/modules/curation_info.rb#L133-L144)).

"Records in curation" are something that was specifically requested before, and they ended up in this part of the dashboard. 
The query spec. can be changed via a server ticket. What would you like me to do with this particular part of the query, @allysonlister, @ramGranell ? Perhaps yet another query/tab could be added only for these records.

ramGranell:
> > In that panel there are records that are approved, e.g. 5237 (I've just checked in the dev server and it has true in approved). I think that the second part of the query adds them ((https://github.com/FAIRsharing/FAIRsharing-API/blob/master/lib/modules/curation_info.rb#L133-L144)).
> 
> "Records in curation" are something that was specifically requested before, and they ended up in this part of the dashboard. The query spec. can be changed via a server ticket. What would you like me to do with this particular part of the query, @allysonlister, @ramGranell ? Perhaps yet another query/tab could be added only for these records.

I don't think that we need these second set of records as they are records that:
1. they are approved 
2. they were assigned to a curator in the curator panel
3. the last editor is a curator (can be a different one that the one appears in 2).
For example, the record https://deploy-preview-2474--fairsharing.netlify.app/5237 was assigned: 1) is approved, 2) was assigned to KayBur and 3) the last registered editor is FAIRsharingTeam (role curator)
@allysonlister what do you think about this?

allysonlister:
I think I agree with @ramGranell here.

What is happening I believe is this:

1. record is updated and goes into the queue.
2. record is assigned to someone eg Delphine
3. record is ultimately approved and we all forget about it
4. some point in the future record is updated again and Delphine remains still assigned.

Curators don't need to remain 'assigned' to a record once it's been approved, though I'm fine if it's easier to keep the code that way. However, we definitely don't need to see records that are assigned but have been approved already - they are finished!!

knirirr:
> However, we definitely don't need to see records that are assigned but have been approved already - they are finished!!

As these are no longer wanted I've created a server ticket: https://github.com/FAIRsharing/FAIRsharing-API/issues/1166
@prakhyatox doesn't need to do anything in this PR in relation to this item; when the server is changed the client will show the different data.

ramGranell (approved):
As one of the features was requested by @delphinedauga , we will need to wait for her approval (not happening this week)

src/components/Curators/CuratorRecordsAwaitingApproval.vue
```diff
@@ -0,0 +1,616 @@
+<template>
+  <v-card class="mb-2">
+    <v-card-text v-if="approvalRequiredProcessed">
+      <v-card-title
+        id="text-curator-search-0"
+        class="green white--text"
+      >
+        <b> RECORDS/CURATOR EDITS AWAITING APPROVAL </b>
```
knirirr:
```suggestion
        <b> CURATOR EDITS AWAITING APPROVAL </b>
```

allysonlister:
> Running the calls individually was not requested in the original ticket and I think is quite complex to change to individual cals (maybe more two tabs with some specific calls). Anyway, I don't think that this is a real problem as how often curators require to enter to this tab, once per month?

No worries - as long as they are only loaded when the tab is clicked, and if @knirirr  is fine with that, I'm also good with it.


However, we now seem to have completely lost the 'least recently updated record', which I found very useful.

> There's not been any change to the query, and it is indeed a bit of a hit.

Perhaps it could be loaded upon a button press?

Also, the tabs haven't been swapped (user edits first) - perhaps I'm just checking the netlify build too soon?

src/components/Curators/CuratorRecordsAwaitingApproval.vue
prakhyatox: Fixed

prakhyatox:
> Also, the tabs haven't been swapped (user edits first) - perhaps I'm just checking the netlify build too soon?
   
Tabs order is updated.

allysonlister:
I'm getting a 500 error on the netlify build.

knirirr:
> I'm getting a 500 error on the netlify build.

That's going to keep happening for a while, until I can work out a way around it.
Meanwhile, I have restarted Nginx.

allysonlister:
Ok let me know when I should check it again. thanks!

knirirr:
> Ok let me know when I should check it again. thanks!

Whenever you like; I have restarted Nginx.

allysonlister:
We still seem to have completely lost the 'least recently updated record', which I found very useful.

> There's not been any change to the query, and it is indeed a bit of a hit.

Perhaps it could be loaded upon a button press?

prakhyatox:
> We still seem to have completely lost the 'least recently updated record', which I found very useful.
> 
> > There's not been any change to the query, and it is indeed a bit of a hit.
> 
> Perhaps it could be loaded upon a button press?

This line appears in just above the table. But the response for the query is bit delay from the data of the table. Therefore, it does not appears as soon as curator dashboard is loaded. If you will wait for a minute the  'least recently updated record' with details will appear

Below is screenshot from netlify build
image

allysonlister:
OK, it's great that it's there, but the issue with the load of the query remains.

@knirirr says there's a reasonably heavy cost associated with it, so I'm suggesting that you don't automaticallly run the query, but rather have a button there that a curator can press if they want the query to run. Does that make sense?

knirirr:
> OK, it's great that it's there, but the issue with the load of the query remains.

I think that splitting it into tabs as has been done will be sufficient. The real problem was always running all of those queries at once.

knirirr (approved):

allysonlister (commented):
thank you!

allysonlister:
I think I approved? It may have done a 'comment' instead, but looks like I approved - thanks!

PR #2480, opened 2025-02-04
Collection text 2472

knirirr:
Some references to recommendations removed from text describing collections.


allysonlister (approved):
thanks!

PR #2486, opened 2025-02-06
Reworking of allowed relations handling #2483.

### LINKED ISSUE #2483: Relationships missing from record edit interface
As you can see from the screenshots below, the toggles for certain registries are completely missing in some of our record edit pages. If you try to search for e.g. a standard to add to a collection, it just doesn't appear; it's like standards are unavailable to collections... 

While it appears to only be a bug for standards, policies and collections, please also check that databases are actually ok and nothing's missing in their relationship edit tabs.

As this is a recent issue, it may be tied to the new FAIRassist registry; note that there should be toggles for FAIRassist components too, right? Certainly they should be accessible via 'related to' from any other record in FAIRsharing, and of course their documented relationships to principles and within metrics and benchmarks.

The following screenshots are what various record types look like in their Relationships edit tabs. Please double check what I have listed as missing against previous versions of the front end to make sure I've got it right. Thanks!

standard: missing databases, collections, policies, FAIRassist
Image

collection: missing standards, policies, FAIRassist
Image

policy: missing standards, collections, FAIRassist
Image

knirirr:
Editing of relations on records should now be correct.
Please take a look at the code (if you can understand it) or try editing some records to check that the correct relations/records are available.


prakhyatox (approved):

ramGranell (commented):
I'm not sure if this is correct, when I'm in a policy, e.g. https://deploy-preview-2486--fairsharing.netlify.app/3397/edit  and I try to create relationship, they appear relations that I think that are not correct, e.g.   a policy extends a collection, a policy is related to a collection, a policy deprecates a collection,  Maybe there are more not correct relations but I didn't check all of them

knirirr:
>  policy extends a collection, a policy is related to a collection, a policy deprecates a collection, 

A policy may link to a collection, but only to recommend it. 

> Maybe there are more not correct relations but I didn't check all of them

Let us hope someone spots them.

ramGranell:
> > policy extends a collection, a policy is related to a collection, a policy deprecates a collection,
> 
> A policy may link to a collection, but only to recommend it.
> 
> > Maybe there are more not correct relations but I didn't check all of them
> 
> Let us hope someone spots them.

For policy, there are the following relationship  extends, is_related, deprecates recommends for four registries (DB, policy, standard and collection). Thing that is incorrect in many of the cases.
For DB,  there are the following relationships implements, accepts, outputs, related_to, shares_code_with, shares_data_with, extends, deprecates for both standards and databases. This is incorrect as from DB to standards only the following relationships are allowed: implements, accepts, outputs, related_to

I can infer the following pattern, if there is a valid relationship between A and B registries then, it puts all possible relationships independently of the registry B, e.g. policy has a valid relation "Policy deprecates Policy" but all the possible relations "Policy deprecates standard", "Policy deprecates collection" and "Policy deprecates databases" appear.

knirirr:
> I can infer the following pattern, if there is a valid relationship between A and B registries then, it puts all possible relationships independently of the registry B, e.g. policy has a valid relation "Policy deprecates Policy" but all the possible relations "Policy deprecates standard", "Policy deprecates collection" and "Policy deprecates databases" appear.

That does appear to be the case.
I have just pushed a further modification which I think fixes that.

ramGranell:
> > I can infer the following pattern, if there is a valid relationship between A and B registries then, it puts all possible relationships independently of the registry B, e.g. policy has a valid relation "Policy deprecates Policy" but all the possible relations "Policy deprecates standard", "Policy deprecates collection" and "Policy deprecates databases" appear.
> 
> That does appear to be the case. I have just pushed a further modification which I think fixes that.

Yes, last push sorted out this. 

I found the following thing:
When you save a new relation, in the existing list of relations (right box) do not appear the icons of the registry status (R,D, U) of the record only appears a ? (it appears normal before saving the new relationships)

src/store/editor.js
```diff
@@ -301,8 +301,20 @@ let editorStore = {
...
+                            // records to show.
+                            if (options.target === null) {
+                                output.push(relation);
+                                seen.push(label);
+                            }
+                            // Whereas this is what determines what available links can be shown when adding a new one.
+                            else {
+                                if (options.target.registry === relation.target ||
+                                    options.target.type === relation.target) {
+                                    output.push(relation);
```
prakhyatox, line 314: This code is same as in line 307.
Is the value of ```relation``` and ```label``` are different in when ```options.target === null``` (line 306) and ```options.target.registry === relation.target || options.target.type === relation.target```(line 312).

It seems different ```if``` conditions but same values are being pushed in array

knirirr: Yes; that code is called in two places:
The first is to populate the list of existing linked records, and the second is to provide a list of allowed links when adding a new one.
It's called with `options.target` as `null` in the first case and as an object in the second. In the first case I need to put everything into the array but in the second then filtering is required. 
This doesn't seem particularly clear but if I'm not keen on refactoring a lot more of the code at this time.

knirirr:
> I found the following thing: When you save a new relation, in the existing list of relations (right box) do not appear the icons of the registry status (R,D, U) of the record only appears a ? (it appears normal before saving the new relationships)


I'm not able to reproduce that locally or on the dev build. 
Do you see any clue in your browser's dev console's network tab?

allysonlister:
IGNORE THIS PLEASE - See comment below at https://github.com/FAIRsharing/fairsharing.github.io/pull/2486#issuecomment-2647441962

I can see the status of a record on the right-hand side.

Everything except one check I've done so far looks ok: but taking the example from my original ticket, this is not resolved yet https://deploy-preview-2486--fairsharing.netlify.app/575/edit:

image

allysonlister:
I can see what Ramon mentions about the "?" - If you use Save and Continue then all right-hand relations statuses are changed to be "?"
image

allysonlister:
Ignore https://deploy-preview-2486--fairsharing.netlify.app/575/edit as this is only meant to have standards and databases in the toggles, this one was my mistake.

ramGranell (approved):

allysonlister (approved):
All good from here, thanks

PR #2489, opened 2025-02-10
Added status to linked record query. #2487.

### LINKED ISSUE #2487: 'Save and Continue' of new relations in any record produces '?' in right-hand relation list
I can see what Ramon mentions about the "?" - If you use Save and Continue then all right-hand relations statuses are changed to be "?"

image

_Originally posted by @allysonlister in https://github.com/FAIRsharing/fairsharing.github.io/issues/2486#issuecomment-2647433709_

knirirr:
This fixes #2487.


allysonlister (approved):
I've only checked this particular behaviour, but it seems to work correctly now. thanks!

ramGranell (approved):
I've tried to create records in three different type of records and this work fine

Issue #2471, opened 2025-01-28
Add a simple pie chart to champions page


allysonlister:
Because https://github.com/FAIRsharing/fairsharing.github.io/issues/2469 is likely to take a long time and is of a low priority level, as a short-term solution perhaps a nice _static_ pie chart as described there could be added to the top half of https://fairsharing.org/community_champions wherever you feel it would best be placed?

Again, the source of the pie chart would be in the intakes spreadsheet, but we aren't looking to show numbers for each wedge, just the research area.

Thanks!

knirirr:
I'll close this as it's covered by #2499.

Issue #2469, opened 2025-01-27
Add subject area filtering for Champions


allysonlister:
Our lovely Champions are getting too numerous to easily see and admire on https://fairsharing.org/community_champions/our_champions and https://fairsharing.org/community_champions/, so here are a couple of modifications that we'd like to implement to make it easier to find Champions based on their broad subject area.

1. Add a pie chart to https://fairsharing.org/community_champions just below the green "meet our community champions" button:
![Image](https://github.com/user-attachments/assets/d14f60ed-506e-4093-b704-154f79e07943)
This pie chart should look a bit like this:
![Image](https://github.com/user-attachments/assets/1d6f69c6-486a-47d5-81ee-3ff39ebdd4f9)
Information to build the pie chart will come from Column F in the "Intakes" spreadsheet, but which can then be propagated to the Champions' json that is used to build our_champions when a new champion is added. This means that the pie chart can be generated on the fly from content in the champions' json info. The pie chart could be clickable and take the user to the our_champions page with the appropriately-clicked filter enabled (see next item).

2. our_champions: Provide tabbed browsing of the Champions based on these same subjects as were used to build the pie chart. I am not sure of the best way to display different configurations of Champions, but please note that this is just one of the upcoming ways we'd like to be able to showcase our Champions. Future ways of viewing champions might include showing e.g. all of the ELIXIR community Champions. This means that whatever solution you provide to show by Subject would also need to be flexible enough to show other qualities of our Champions. Maybe something like this badly-drawn mockup? Here, I would only want the Research Area to be implemented right now, but it gives us scope to have more at another time:
![Image](https://github.com/user-attachments/assets/26698a72-ef17-4eeb-9683-4e163c0ed8dd)

Other small updates to community_champions:

1. Change "network of experts available on our Community Champions Slack channel" -> "network of experts available on our Community Champions Zulip channel"
2. Change "Our champions are properly cited" -> "Our Champions are properly cited"
3. Change "Community Champions have their own Slack workspace" -> "Community Champions have their own Zulip workspace"
4. Change "Community champions help their own research domains by creating records" -> "Community Champions help their own research domains by creating records"

allysonlister:
Or, if it can be incorporated into the "Year" pulldown, that would be great too, whatever's easiest for you.

PR #2499, opened 2025-02-18
Champion pie chart

knirirr:
This should satisfy #2471 and act as a starting point towards #2469.


ramGranell (approved):
The pie chart looks fine for me. At the moment is not clickable (e.g. chemistry sector goes to chemistry Champions) and I don't know if it possible to change it. Anyway, this is the second point of the ticket https://github.com/FAIRsharing/fairsharing.github.io/issues/2469

knirirr:
> At the moment is not clickable

That's right. Linking to separate tabs per specialism will be a separate ticket.

allysonlister (approved):
looks really great, thank you! (I note that you say making it clickable is another ticket, thanks!)

Issue #2500, opened 2025-02-20
Graph relations


knirirr:
Some new relations have been added and these will need to go on the legend to the left of the graph page. 
The relations have been added to the graph in this PR: https://github.com/FAIRsharing/FAIRsharing-API/pull/1174

Also, new registries etc. will be needed.

knirirr:
These have been deployed.

PR #2502, opened 2025-02-20
New graph relation labels 2500

knirirr:
This deals with #2500 

Important note: Graphs may be incorrect until a full rebuild is done on the development server using this branch: https://github.com/FAIRsharing/FAIRsharing-API/pull/1174


ramGranell (commented):
It is not easy to review this one as there is not a Netlify to check the graph.

knirirr:
Not much to be done about that other than to look at the source (which needs to be done anyway) or run it locally, I'm afraid.

allysonlister (approved):
From looking at the code differences, this seems ok, though I can't look at the graphs, as Ramon says. thanks!

PR #2525, opened 2025-03-25
New curation requirements 2518

knirirr:
This changes the message users see after creating a record, and also adds a new tab for incomplete records (may hog RAM) on the curator dashboard.


ramGranell (commented):
Various things to comment here:
1. About the functionality of what was requested:
I think it works correctly. The only doubt is what happened when a record is created as a curator. I don't have clear if it is approved or it isn't.

Now that we have a list of mandatory fields, I think that we need to remove the server automatic check of the TAXONOMY field as the error shown is not clear/intuitive and not a very nice format. In addition why TAXONOMY has to have a specific server check such as record name and not other field such as SUBJECT


2. About the intuitive the process is. 
I found that this is not really intuitive. Finding the fields that are missing is not easy based on the documentation. I think we need to indicate better where each one of the fields can be updated. Ways to do this can be: 1) indicating in the email step by step how to do the things, e.g. a) login in FAIRsharing, b) go to your record, c) go to the edit record panel, in the right menu called ACTION select EDIT c) at the top of the edit record panel there are X tabs, go to tab Y and do Z , 2) including a short video (gif) indicating how to do this in the documentation (if it is possible) or 3) including images with the steps one by one

This is my experience of an important change, maybe we need the real experience by users that are not so familiar with FAIRsharing. Why can be tested by the Champions in the Oxford meeting?

ramGranell:
When I go to incomplete record, it starts loading and never ends. In console there is the following error:
```
vue.runtime.esm.js:3065 TypeError: Cannot read properties of undefined (reading 'forEach')
    at IncompleteRecords.vue:121:1
    at Array.forEach (<anonymous>)
    at s.prepareIncompleteRecords (IncompleteRecords.vue:119:1)
    at s.mounted (IncompleteRecords.vue:108:1)
```

knirirr:
@ramGranell what steps are you following to get to an incomplete record? What page are you on when you see the console error?

ramGranell:
> * netlify/fairsharing/deploy-preview

I log in https://deploy-preview-2525--fairsharing.netlify.app/curator
I go to curator panel
I go to incomplete record tabs
It start loading and never finished
it said:
Failed to load resource: the server responded with a status of 504

knirirr:
Thanks. The required data appear to have been missing, so I have run `RAILS_ENV=production rake modifications:check_completeness` in order to fix it.

PR #2530, opened 2025-03-31
Add codespell support (config, workflow to detect/not fix) and make it fix few typos

yarikoptic:
More about codespell: https://github.com/codespell-project/codespell .

I personally introduced it to dozens if not hundreds of projects already and so far only positive feedback.

CI workflow has 'permissions' set only to 'read' so also should be safe.

Note that some fixes could be (positively or negatively) affecting functioning -- seems found typos in HTML elements and some test and fixtures -- please review carefully.


.github/workflows/codespell.yml
```diff
@@ -0,0 +1,25 @@
+# Codespell configuration is within .codespellrc
+---
+name: Codespell
+
+on:
+  push:
+    branches: [master]
+  pull_request:
+    branches: [master]
```
knirirr:
```suggestion
  [push, pull_request]
```

knirirr (approved):

.github/workflows/codespell.yml
yarikoptic: we could do that, but likely would need to get to the same line as `on: `?

yarikoptic: but also it would lead to double runs I think -- it might then run for the PR branch (e.g. "blah") and also then for the PR itself... adding noise etc.  So -- are you sure? what are you trying to achieve with widening the scope of when to run?

yarikoptic:
oh, note to myself -- I have not sinned, it was a CI which triggered the documentation regeneration and fully pretended to be me ... but badly. Filed:
- https://github.com/elstudio/actions-js-build/issues/20

.github/workflows/codespell.yml
knirirr: I would want tests to run when we make PRs from feature branches to our dev branch (we don't merge features directly to master) or between our feature branches.

yarikoptic: so let's add `dev` there?
```suggestion
  push:
    branches: [master, dev]
  pull_request:
    branches: [master, dev]
```
so that pushes and PRs against `dev` are also tested?

knirirr: That wouldn't include feature->feature PRs.

yarikoptic: oh, I have missed the "or between our feature branches" part... ok -- your wish my act ... although I think this might be too much and you might want later to make it for pushes only to `master` and `dev` and all other PRs.

yarikoptic:
but let me redo this PR which apparently had a wrong base...

PR #2541, opened 2025-04-03
Refactoring to solve #2512

### LINKED ISSUE #2512: Make reference URL hyperlinked at bottom of record
Reference URLs (fields available in collections and in Principles) are already clickable within the General Information card at the top of a record, but are not clickable at the very bottom of the record where they are also present.

As a low priority ticket, please can you make the reference URL clickable at the bottom of the page too?

Thanks!

knirirr:
This fixes #2512 by using a regex to recognise URLs rather than looking for the hardcoded meadata key name.


src/components/Records/Record/AdditionalInfo/DatasetArray.vue
```diff
@@ -151,6 +137,11 @@ export default {
         }
       }
       return false;
+    },
+    checkRegex(string) {
```
prakhyatox, line 143: Could you please add a comment about what this regex is checking, that would be helpful

knirirr: No problem.

prakhyatox (approved):

Issue #2520, opened 2025-03-20
Adv Search: 404 error


allysonlister:
This is actually a slightly incorrect query, but there's no reason it shouldn't work, yet it gives a 404. Does it do the same for you?

![Image](https://github.com/user-attachments/assets/1adf946e-b379-4f64-ac26-d73cbacd590b)

https://fairsharing.org/advancedsearch?q=ISO&operator=_and&fields=%28operator%3D_and%26registry%3Dstandard%26status%3Dready%26organisations%3DInternational%20Organization%20for%20Standardization%20%28ISO%29%29

allysonlister:
Refreshing the above search produces an empty search result rather than a 404 - perhaps it has something to do with the parentheses in the org name?

knirirr:
It loads first time for me (0 results though).

allysonlister:
It is now loading for me first time, but there shouldn't be zero results, either with that query or this simpler one: https://fairsharing.org/advancedsearch?operator=_and&fields=%28operator%3D_and%26registry%3Dstandard%26status%3Dready%26organisations%3DInternational%20Organization%20for%20Standardization%20ISO%29

You can confirm this with the ISO org page: https://fairsharing.org/organisations/4125

So there is a bug there somewhere...

knirirr:
> Refreshing the above search produces an empty search result rather than a 404 - perhaps it has something to do with the parentheses in the org name?

That's exactly what it appears to be. They're stripped out of the query sent to the server and since that's looking for exact matches to organisation names it doesn't find them. The query works if I put the brackets back.
The question is now how I might ensure that they are sent to the server in the first place...

knirirr:
The problem is located here: https://github.com/FAIRsharing/fairsharing.github.io/blob/master/src/views/AdvancedSearch/AdvancedSearchResultTable.vue#L233-L235

@prakhyatox do you have any thoughts on how this might be resolved?

prakhyatox:
I will have to look into it.

PR #2545, opened 2025-04-07
Fix: allowing search with values having parenthesis ()

prakhyatox:
Ticket: https://github.com/FAIRsharing/fairsharing.github.io/issues/2520

The fix is also implemented in on going vue3 migration


knirirr:
I had a look at this now and I found that the parentheses were still being stripped:

Screenshot From 2025-04-07 16-25-09

...which presumably the causes the query to fail as exact matches are needed for faceting. 
If I then edit the query and select the correct organisation (a spurious one without the parentheses appears in the list) it works the first time, but strips them again if the URL is re-used.

prakhyatox:
> I had a look at this now and I found that the parentheses were still being stripped:
> 
> Screenshot From 2025-04-07 16-25-09
> 
> ...which presumably the causes the query to fail as exact matches are needed for faceting. If I then edit the query and select the correct organisation (a spurious one without the parentheses appears in the list) it works the first time, but strips them again if the URL is re-used.

Fixed

knirirr (approved):

Issue #2555, opened 2025-04-30
FAIRassist creation page


knirirr:
To be available at fairsharing.org/create-fairassist.
As per the current creation page, but with the available types constrained to only FAIRassist and a new title (e.g. "Creating a new FAIRassist record"). 
Probably the best way to do this would be to use the existing page but pass props from the router depending on the route called, e.g. ` props: { fairassistOnly: true }`

PR #2556, opened 2025-04-30
Creates a separate /create-fairassist route. #2555

knirirr:
Now, there's a /create-fairassist link (not linked from anywhere at present) which will make changes to the creation form as per #2555.


allysonlister (approved):
looks good to me, thanks!

prakhyatox (approved):

Issue #2557, opened 2025-04-30
quick links for FAIR Principles in a Nutshell


allysonlister:
I can't get the following two URLs to work - when I load them, even refreshing, the little tile doesn't appear 'in front of' the educational page. Please could someone also check this is the case for them? thanks!

> Oh, and please could the quicklink be https://fairsharing.org/educational#fair and https://fairsharing.org/educational#FAIR ?
> 
> 
> 
> thanks! 

 _Originally posted by @allysonlister in [#2351](https://github.com/FAIRsharing/fairsharing.github.io/issues/2351#issuecomment-2178885258)_

knirirr:
It works for me using the #fair URL but not #FAIR. 
I can investigate this case-sensitivity issue but I am not sure what is going on in your browser. Presumably Firefox (which I also use)?

allysonlister:
I presume it was some browser nonsense. When I try now, i have the same behaviour as you.

knirirr:
This will be deployed next time there's a merge to master.

PR #2559, opened 2025-05-07
Made educational page fragments case insensitive. #2557

knirirr:
This fixes the issue in #2557.
It would make sense to merge this into master along with the other small bugfixes which are still queued up.


allysonlister (approved):
working as expected thanks!

Issue #2565, opened 2025-05-13
"Awaiting DOI" message update for non-ready records


allysonlister:
![Image](https://github.com/user-attachments/assets/7a068243-7dc4-4dbc-9131-f2500db84502)

After approving https://github.com/FAIRsharing/FAIRsharing-API/issues/1234, I realised that it would be good to have a different message for non-ready records that do not have a DOI.

Ready - keep as-is, with "Awaiting DOI" hyperlinked to https://fairsharing.gitbook.io/fairsharing/#getting-a-record-doi

Uncertain, Deprecated + does not yet have a DOI (rare, but will happen) - "DOI will not be issued", and hyperlinked to the same gitbook section of https://fairsharing.gitbook.io/fairsharing/#getting-a-record-doi

In development - "DOIs are only issued to records with a 'Ready' status", and hyperlinked to the same gitbook section of https://fairsharing.gitbook.io/fairsharing/#getting-a-record-doi

Thanks!
Allyson

PR #2567, opened 2025-05-14
Changed awaiting doi message. #2565

knirirr:
After a looking at the ticket for a few minutes I realised that this one didn't need much work, so I've done it now, before I start on something more complex.


allysonlister (approved):
looks great, thank you!

PR #2606, opened 2025-06-20
Added a delete button to the incomplete records curator tab #2585.

### LINKED ISSUE #2585: Delete record button in the incomplete record tab in the curator dashboard
As it happened in other tabs, a button to delete records is required (a button to accept records is not required here). This allow the curators to reject the records that are clearly not feasible without need to wait the cleaning period process.

knirirr:
A delete button has been added as requested in #2585.
N.B. some records were taking a long time to delete, for reasons that aren't clear. Many record associations were being checked, even if the record appeared to have few links. 
Is this happening elsewhere in the curator tab?
Anyway, it's not related to this Javascript code and so can be investigated separately if needed.


ramGranell (commented):
When trying to delete the record 4738, there is this error
Screenshot_20250623_082932
and it was not able to delete it

allysonlister:
image
I can see the delete record column, but the page itself is taking a very long time to load (a few minutes) and hasn't finished loading yet. I'll let you know if i can test it or not.

allysonlister:
It has not loaded. I will continue trying.

allysonlister:
managed to get in

allysonlister:
I can confirm that I can see and press delete, but it is taking much longer than it does in the "user edits awaiting approval" - perhaps it's trying to regenerate the "incomplete records" tab contents?

In fact it's taking so long that I fear it may be hanging, but I will wait and see if it ultimately works or if it throws an error.

ramGranell:
To see the error that I commented before, you need to open the inspect option in the browser (it is in the right button menu)-> console tab

allysonlister:
Mine is still stuck on the deleting and apparently doing nothing?

image

allysonlister:
still going... Not sure how I can test this.

allysonlister:
I have refreshed the curator panel and am attempting to re-delete. @knirirr I'm going to have to recuse myself from this review if I cannot even test the behaviour.

image

allysonlister:
The delete popup definitely just hangs for me when I try it - I have given it lots of time and tried it twice. Not sure what else I can do at this stage, sorry!

knirirr:
This code is deleting records successfully, though it is taking a while to do it (30 seconds, perhaps). The reason for the long time (other than the usual curator dashboard issue) is this:

https://github.com/FAIRsharing/FAIRsharing-API/blob/master/app/models/fairsharing_record.rb#L20-L24

...i.e. callbacks which run on every delete action. I have spent some time looking at those and can't think of anything which will speed them up, though I have made one small change to them.

One other issue I discovered is that Neo4j had broken (related to security settings for the APOC plugin), and if any database connected to the server shuts down the records won't save. I have now fixed Neo4j.

ramGranell (approved):
It takes around 30 sec but it seems to work properly for me.

allysonlister (changes requested):
For whatever reason it is now allowing me to delete, so the behaviour is great thanks.
I note however that this delete popup does not include the name or abbrev of the record as our other delete popup does. This is a valuable final check. Might we be able to include this? thanks!

knirirr:
The data available to this page are organised differently from the other page with a delete button. I can show the ID in the dialog (and have done so) but if that is insufficient it will take a bit longer

allysonlister (approved):
great, an id is super. thanks!

knirirr:
Thanks. I will deal with this later (though https://github.com/FAIRsharing/FAIRsharing-API/pull/1304) still needs looking at.

Issue #2610, opened 2025-06-23
FAIRsharing Accessibility: show record status


prakhyatox:
User's description: 
User Name: Baptiste Cecconi 
> May I ask that we have a color-blind-friendly theme option on FAIRsharing. I'm specifically thinking of the "User private profile page", in the Record listing tabs: the dots symbolizing the approval status are really difficult to read for me. There are colorblind friendly color palettes, but more easily, if there is a double encoding (color, if you wish to keep, plus a symbol), then it makes it accessible for all. Colorblind people are ~10% of us (~10% of men, ~0.5% of women).

[#Community Champions > FAIRsharing Accessibility @ 💬](https://zulip.fairsharing.org/#narrow/channel/20-Community-Champions/topic/FAIRsharing.20Accessibility/near/63409)

PR #2611, opened 2025-06-23
Fai rsharing accessibility show record status 2610

prakhyatox:
Ticket: https://github.com/FAIRsharing/fairsharing.github.io/issues/2610

image


allysonlister (changes requested):
looking good! 

Please could you update the maintenance request panel (approved/not approved)?
image

I note that the subject browser has also been done, but for statuses - thanks! This needs no further modifications
image

allysonlister (approved):
looks good thanks!

PR #2614, opened 2025-06-27
Fix for #2163

### LINKED ISSUE #2163: Remove extraneous full stops from related records

knirirr:
This fixes #2163.

It appears that the response from the server is `[{message: 'record not found'}]` if there's no graph.


prakhyatox:
While trying to replicate the error scenario in local, I am getting below console error

Screenshot 2025-06-27 at 11 50 01

knirirr:
I was able to reproduce that on a different browser. Please try it now.

prakhyatox (approved):

PR #2622, opened 2025-07-07
Changed text #2621.

### LINKED ISSUE #2621: Renaming 'Object Types' in record display only
Hi all,

It seems to be confusing for some users to see only "Object Types". Particularly for non-database records, just having "Object Types" seems to make the user think they should describe the resource itself rather than its scope.

Please could you change "Object Types" just in the display of the record's General Information card to 

"Object Types in scope for this resource"

It is 4 characters longer than "Countries developing this resource", but hopefully will be ok for the display.

Thanks!
Allyson

knirirr:
Text changed as requested.


allysonlister (approved):
Great, thank you!

Issue #2628, opened 2025-07-23
New banners for incomplete records


ramGranell:
There are three cases with incomplete/awaiting review records: 

1) record is awaiting review and it is complete (current banner),  
2) record is awaiting review and it is incomplete (banner saying "This record has been recently modified, but it is NOT be sent to the curation team until all requirements are met. More information is available in our [documentation](https://fairsharing.gitbook.io/fairsharing/record-sections-and-fields/how-to-update-a-record#minimal-curation-requirements)." (I made your message short) 
3) record is not awaiting review and is incomplete and is not deprecated,  "This record does NOT meet all the minimal curation requirements. More information is available in our [documentation](https://fairsharing.gitbook.io/fairsharing/record-sections-and-fields/how-to-update-a-record#minimal-curation-requirements)."

This ticket may require a server ticket if the field incomplete is not accessible by the client. Also, it can be modified if it is very different to distinguish among the different cases.

allysonlister:
Thanks for explaining all of this in the ticket.

Please can you modify 2 to "This record has been recently modified, but will not be sent to the curation team until all requirements are met. More information is available in our [documentation](https://fairsharing.gitbook.io/fairsharing/record-sections-and-fields/how-to-update-a-record#minimal-curation-requirements)."

I'm hesitant to implement 3 until we have at least gotten the db and std spreadsheet pushed to production, as otherwise it may confuse users.

So 1 and 2 for now, then 3 into another ticket to be done once David's (and the rest of the team's) new curation is pushed to production. Does that sound good?

Thanks!

knirirr:
Part two is done and will be deployed shortly.
I will leave this open for part three.

knirirr:
@allysonlister @ramGranell IIRC quite a bit of curation has been done since this ticket was opened, so are we ready for part 3 yet?

allysonlister:
I think we need to wait until at least the standards curation spreadsheet has been pushed to production, as there will be many that will improve once their object types are added.

PR #2629, opened 2025-07-23
New banners 2628

knirirr:
Changing banner text as per #2628 part 2.


allysonlister (approved):
working as expected, thanks!

PR #2635, opened 2025-09-05
Remove social login 2634

knirirr:
Social logins should now be hidden on development builds but not on a local development machine. 
This is because the logins won't work there - Github and ORCID aren't configured to use them, and I don't think I can change that.


allysonlister:
I'm not sure I can accurately review this... What would you like me to check?

knirirr:
What to check is that you're happy with this change and the ORCID etc. buttons do indeed not appear when you look at the development build.

allysonlister:
So https://deploy-preview-2635--fairsharing.netlify.app/ is behaving as it would on eg preview?

allysonlister:
Thanks. The buttons have disappeared, as you have said, so that looks good. However, I have two suggestions:

1. perhaps we need to modify the link underneath the login window:

image

Instead of "Can't login with ORCID?" perhaps now the link text should be "Can't login with ORCID or GitHub?" 

2. I suggest you add more text below your new text "Please note that ORCID and GitHub authentication will not be available if you're using one of our development systems; they are only set up to work on the main fairsharing.org site." in https://fairsharing.gitbook.io/fairsharing/#accessing-fairsharing-through-3rd-party-accounts. Such text should remind users that, if they need to get a password for their TPA account, they need to request a password reset and then they can log in to the dev server without their TPA.

knirirr:
I didn't add any explanatory text there as I think users have trouble reading it and users logging into the dev site is going to be unusual anyway; I put it on Gitbook instead (there's a pending merge request IIRC). I could add "..or Github", though.

allysonlister (changes requested):
Apologies I wasn't clear. as per my item 1, adding "or Github" is what I'd like to see thanks. 

as per item 2, I meant updating the Gitbook documentation underneath your new change, which I have already approved.

allysonlister (approved):
thank you!

Issue #2658, opened 2025-11-10
New tool to add to communities#tools


allysonlister:
Hi all,

there is a new google form response for the tools that integrate FAIRsharing. Please let me know if you don't have access to the form, otherwise please can you add it to our webpage? https://fairsharing.org/communities#tools

thanks!

PR #2659, opened 2025-11-11
Added new tool #2658.

knirirr:
A Netlify build may not appear for this PR but the code is simple enough to look at. 
We don't have a record for the organisation supplied so there were two choices:

1. Create one.
2. Change the code to link to the organisation's homepage rather than ours.

I chose (2) because I'm not certain if their organisation will ever be linked to a fairsharing_record, so there's a chance it may be purged.


prakhyatox (approved):

allysonlister (approved):
looks great, thank you!

Issue #2673, opened 2025-12-02
Advanced Search including the new id schema fields (globally unique, persistent, resolvable)


allysonlister:
If the vue3 upgrade for the advanced search is complete and it won't take much of @prakhyatox's time, then in the new year please could he add a new rule to the advanced search within the standards pulldown that allows users to select these three fields. I appreciate that these fields only exist within id schema record types, but it should be ok.

If this needs to be put on hold in favour of vue3 upgrade, then that's fine too; I'll let you know if it gets more urgent.

PR #2675, opened 2025-12-03
New search fields 2673

knirirr:
This deals with #2673.
As far as I can tell the added files should be OK when upgraded to Vue3.

Looks like reindexing the dev server is necessary. I'm doing that now.


allysonlister (commented):
Ah did I not notice the typo where it said 'globally unique' too much?
image


Also, can I ignore this:
image

I can't tell from the code, but will these options allow a multi-select in the advanced search (e.g., select both persistent and resolvable?)

allysonlister:
(I'll check the netlify build)

allysonlister:
Ah I see now - they are radio buttons as individual elements - that's great!

allysonlister (approved):

ramGranell (approved):
The search works properly, my only suggestion is to change the font size of the chips or the background colour as I think they are not easy to read (please compare them with the other chips of subjects, domains and taxonomy)

knirirr:
Thanks. I picked that size specifically (being aware of other chips) as I have no problems with readability and prefer them to fit into the line of text. 
How desperate are you for them to be changed? Do you have any thoughts @allysonlister ?

ramGranell:
> Thanks. I picked that size specifically (being aware of other chips) as I have no problems with readability and prefer them to fit into the line of text. How desperate are you for them to be changed? Do you have any thoughts @allysonlister ?

In the previous version there was an upper-case text "GUPRI" but now Ihave problems to read the lower-case white font over light blue in a small font. Anyway, it would be great to ask an expert of user interface.

knirirr:
@ramGranell in that case I suggest that you create a PR adjusting these labels to your preference; we can then be sure that you find them sufficiently easy to read.

allysonlister:
Please can you use this order for the three fields in the edit screen and the view screen?

1. globally unique
2. persistent
3. resolvable

PR #2686, opened 2026-01-05
modifying 2026 Champions

ramGranell:


allysonlister:
Alex Rooney is missing, please could you add her? She does have 2026 marked in the Intakes sheet.

FYI Kyle, Laurence and Kay haven't replied, so it's correct to show them as alumni, but we may need to re-add them if they do get back to us.

allysonlister (approved):

ramGranell:
I added A. Rooney

Issue #2695, opened 2026-01-22
Updating educational factsheet DOIs and download links to latest version


allysonlister:
I've noticed that some of the links on https://fairsharing.org/educational are going to older versions, eg policies 

<img width="397" height="408" alt="Image" src="https://github.com/user-attachments/assets/f4c61154-c19b-45b4-bc3e-d5aef8c96199" />

Please can we make sure the direct download link for each is going to the most recent version, and that the copyable link goes to the "cite all versions" doi, e.g.

<img width="572" height="111" alt="Image" src="https://github.com/user-attachments/assets/4789de87-9f99-478c-8e5f-4689291efa1d" />

Thanks!

knirirr:
It will be necessary for a ticket to be opened every time one of these sheets is changed if there is to be a hard-coded direct download link to the latest version. This could be avoided by only linking to e.g. https://doi.org/10.5281/zenodo.11059818 rather than directly to the PDF, as we originally had it. This would be my personal preference anyway as I don't like to open a PDF in the browser without prior warning.

allysonlister:
I agree completely that it would be preferable to not link to the PDF, but that was a requested feature. I'll let you know if it changes in future, and I shall definitely make sure that, in future, I make a new ticket each time we update one.

PR #2697, opened 2026-01-22
Modified DOIs for infographics. #2695.

knirirr:
I've changed the DOIs to the "cite all" version in each case. As far as I can tell all the PDF links are correct except for policies overview, where a v. 1.1 is listed but there appear only to be v. 1.0 documents, as far as I can see. Could the v. 1.1 have been mislabelled as 1.0?


allysonlister:
Thanks for doing all of that, looks great.

v1.1 of the policies contains an unmodified v1.0 of the english lang version, and v1.0 of the new portuguese translation.

So it's a new version of the zenodo record (1.1) that newly incorporates a v1.0 portuguese version but an unmodified english one. Does that make sense?

If you have a clearer way of versioning I am happy to amend the record...

allysonlister (approved):

knirirr:
I think from what you have described that I've linked to the correct file - thanks.

Issue #2700, opened 2026-01-23
Record claim login link contains extraneous text


knirirr:
A user attempting to claim https://fairsharing.org/10.25504/FAIRsharing.1ky0cs was taken to an invalid link of https://fairsharing.org/10.25504/accounts/login?goTo=%2FFAIRsharing.1ky0cs. Evidently the code generating that link was expecting users to visit  https://fairsharing.org/FAIRsharing.1ky0cs instead.

ramGranell:
please, tell me when the ticket is done to email to the user that discovered the bug.

PR #2703, opened 2026-01-26
Fixed link. #2700

knirirr:
This generates the link in a manner which avoids the DOI problem encountered in #2700.


ramGranell (approved):
In the Netlify, when I'm either https://deploy-preview-2703--fairsharing.netlify.app/2502 or https://deploy-preview-2703--fairsharing.netlify.app/FAIRsharing.1ky0cs and I try to claim the record, it tries to go to
https://preview.fairsharing.org/accounts/login?goTo=%2F2502 and https://preview.fairsharing.org/accounts/login?goTo=%2FFAIRsharing.1ky0cs respectively
and it ends here: https://preview.fairsharing.org/ 

I suppose that it would be Netlify issuess

knirirr:
That's because of the Netlify configuration. The URL will redirect to the correct server if run locally or in production. 
The important part is that it removes the extraneous part of the DOI from the link when users visit a record with its full DOI inc. number.

Issue #2699, opened 2026-01-23
Invalid links on Negative/Positive/Indeterminate examples fields


rduyme:
**Describe the bug**
Invalid links on Negative/Positive/Indeterminate Examples for use case :  text + URL in Note field

**To Reproduce**
https://fairsharing.org/7494

**Expected behavior**
no 404

**Screenshots**
<img width="1292" height="519" alt="Image" src="https://github.com/user-attachments/assets/39cc36d9-8a6f-408f-8eee-8888635a6a40" />

knirirr:
It appears that what's happened here is that someone has edited a record to add a string with http in it to a notes field, and the system has identified the field as a URL, which it isn't because it contains extraneous stuff. The fix is to not turn that text into a hyperlink.

PR #2704, opened 2026-01-26
"Fixed" URL problems #2699.

knirirr:
This will deal with #2699, at the risk of XSS issues. But, if we don't convert text into links then we're bound to get a ticket about links not being clickable. What are your thoughts, @prakhyatox ?


prakhyatox (changes requested):
Adding new tag with URL in the name, the text is still converting to URL

Image

knirirr:
> After adding new tag, still text is converting to URL. 

That's what's supposed to happen. I changed the code from the old version of "if there's http in the code, make the whole thing into a link" to "if there's a valid URL in the text turn only that into a link". To see this, try e.g. 

"This is the most awesome website: https://fairsharing.org"

...in one of these fields. 
The question is then: Is it worse to have this with its potential risks or have users complain they "can't click on links"?

prakhyatox (approved):

PR #2715, opened 2026-02-05
Dbtobench 2714

knirirr:
This allows databases to be related_to benchmarks.


allysonlister (changes requested):
If I try to remove the relationship from this benchmark (https://deploy-preview-2715--fairsharing.netlify.app/7162) to the FS record, then I get a 500 error. 

Note this is an 'old' relationship that was already in the dev db going from the benchmark to the FS record. 

I was trying to delete it to then move to the FS record itself and create a new related_to relationship from the FS record to the benchmark in order to test it.

knirirr:
This is unrelated to the PR; Neo4j had stopped running on the development server.

allysonlister (approved):
thanks looks good!

Issue #2717, opened 2026-02-12
Reminding users about missing fields


knirirr:
Sometimes, users appear to not notice or understand the email summaries explaining what fields they need to add to complete their records. A way to remind them could be as follows:

A link at the top banner could say something along the lines of "Record incomplete. No DOI will be issued. Click here to see what's missing." This would open a pop-up field showing any missing required fields, perhaps a copy button, and links to the Gitbook docs. These data are all available from the server.

PR #2723, opened 2026-02-19
Added an alert for missing required fields. #2717

knirirr:
This deals with #2717 in a relatively easy way, without having to make any significant change to the existing alert structure. 
The missing required fields alert will be visible to anyone, whether logged in or not, my reasoning being that it might be of interest to users in general to know that a record is missing some information. It could (probably, not necessarily trivially) be changed to be shown only to the record's maintainer, or to curators, or logged-in users etc. etc.


allysonlister:
Looks good. How difficult would it be to also add recommended (but not optional) items please?

knirirr:
> Looks good. How difficult would it be to also add recommended (but not optional) items please?

Easy enough, but the more I cram in an alert the more cluttered it would look.

allysonlister:
Please could you try the recommended on a newline within the banner (or however you think the crammed in look will look best)? Often eg the policy attributes and db attributes are only recommended and we have to prompt people (bc it's hard to check for completeness when there is a default value).

allysonlister (changes requested):
Looking good - i think we should go with the two banners.

However, note that some records eg this deprecated one https://deploy-preview-2723--fairsharing.netlify.app/FAIRsharing.rnx28e and this 'ready' one https://deploy-preview-2723--fairsharing.netlify.app/2086 say "will not be issued a DOI" when they already have a DOI. Can there be a check for DOIs already in incomplete records, and then a different message used please?

allysonlister (approved):
thanks all good!

prakhyatox (approved):

PR #2732, opened 2026-02-27
Only show incompleteness banners to curators and maintainers. #2730.

### LINKED ISSUE #2730: New blue banner listing missing fields: change visibility
Hi all,

The banners look great and are really helpful. However, I'd like to modify who sees them and under what conditions please. Now that I'm looking at them regularly (eg they apply to almost all standards as we haven't pushed the standards obj type updates yet) I think it might give people too much to look at.

Curators and above: always see all of them
Maintainers: see all of them when logged in
Everyone else: banners are not visible.

Once we push the standards curation, we may change the "everyone else" from banners are not visible to required fields banner is visible.

Thank you!

knirirr:
This should now only be visible to curators and maintainers.


allysonlister (approved):
looks good thanks!

prakhyatox (approved):

Issue #2762, opened 2026-05-07
Advanced search underscores


knirirr:
An example has been reported where record type has been sent by the advanced search client as "identifier schema" instead of "identifier_schema", which results in no hits.

PR #2768, opened 2026-05-12
Advanced search underscore bugfix 2762

prakhyatox:
Ticket: 

1. https://github.com/FAIRsharing/fairsharing.github.io/issues/2762
2. https://github.com/FAIRsharing/fairsharing.github.io/issues/2769


allysonlister (changes requested):
re: advanced search

I can only test those queries where I know the problem existed, which is just those with identifier_schema. That has worked fine, but please do double check there aren't any other names/labels with underscores that have the same problem.

re: external (in this case cross reference) links opening in a new window:

While the cross references now open in a  new window, I noticed by chance that the ORCID URL does not, see 
Image

Please can you check all external links in a record and find if there are any other cases where the links do not open in a new window as per most of the existing external link fields?

knirirr (approved):

prakhyatox:
> re: advanced search
> 
> I can only test those queries where I know the problem existed, which is just those with identifier_schema. That has worked fine, but please do double check there aren't any other names/labels with underscores that have the same problem.
> 
> re: external (in this case cross reference) links opening in a new window:
> 
> While the cross references now open in a new window, I noticed by chance that the ORCID URL does not, see Image
> 
> Please can you check all external links in a record and find if there are any other cases where the links do not open in a new window as per most of the existing external link fields?

To check all the links, it should be in a separate ticket. As, checking each link will take time. It would be helpful if some-one can do a round of testing to check all the links.
If both the tickets has been resolved in this PR, then I think we can merge this PR.

knirirr:
> To check all the links, it should be in a separate ticket.  .... If both the tickets has been resolved in this PR, then I think we can merge this PR.

I agree; it would be good to get this fix deployed ASAP.

allysonlister (approved):
Happy to push checking of other underscores to another ticket

PR #2778, opened 2026-05-19
Assigning a curator should not approve a record. #2777

### LINKED ISSUE #2777: Curator assignment; skip_approval
Based on: https://github.com/FAIRsharing/FAIRsharing-API/issues/1584

It appears that the client has changed and is no longer sending `skip_approval: true` when assigning a curator to a record.

knirirr:
This should resolve #2777.


allysonlister (approved):
I can't see a netlify build but from how you described the issue, the proposed modification seems good to me. thanks!

PR #2784, opened 2026-05-22
Replacing prerender 2754

prakhyatox:
Ticket: https://github.com/FAIRsharing/fairsharing.github.io/issues/2754

This PR includes changes which implements SEO tags.

SEO tags are skipped in Netlify URL, it is unnecessary to have longer build time.
SEO tags and site behaviour can be checked on [testing.fairsharing.org](https://testing.fairsharing.org), which mirrors production.

To test the SEO tags, right click on the page and select ```View Page Source``` option.
Scroll down to check the title tag and description tag. Each page will have it unique title and description. If not then it show default FAIRshaing title and its description

Netlify URL: https://deploy-preview-2784--fairsharing.netlify.app/


David-Tomkins:
Hi Prakhyat - do I need access to https://testing.fairsharing.org/ for this as it comes up 403 Forbidden...  Thanks, D.
________________________________

prakhyatox:
@David-Tomkins Please try now. There was build running earlier but should be working now

knirirr (approved):

ramGranell (approved):
I've checked various records and they look ok. Question what happens if there are two records with the same name?

David-Tomkins:
Yup, working now!  Thanks.
________________________________

prakhyatox:
> I've checked various records and they look ok. Question what happens if there are two records with the same name?

Do you have any example?

ramGranell:
> > I've checked various records and they look ok. Question what happens if there are two records with the same name?
> 
> Do you have any example?

No really, but you can create two records with the same name and check what happens

prakhyatox:
> > > I've checked various records and they look ok. Question what happens if there are two records with the same name?
> > 
> > 
> > Do you have any example?
> 
> No really, but you can create two records with the same name and check what happens

Whatever the name of the record it, it will take it title as its name. It is only for SEO, so the index.html files are generated for the pages.

David-Tomkins:
I've done a fair few spot-checks and seems fine - there does seem to be some unnecessary repetition in the record at https://testing.fairsharing.org/FAIRsharing.af0c1d but that may be because of the weird formatting (which I'd not come across before)?

PR #2828, opened 2026-07-09
Prerender script fix

knirirr:
@prakhyatox I've had a lot of trouble with this script. It's been keeping values in a cache somewhere so that changes are not reflected (which you found with a recent community fix) and changes to .env also aren't noticed.
File permissions were also preventing some images from showing. 
A plain `vike build` works, so I have tried simplifying the script.
It should not be necessary to fetch data from the API on a production build as current JSON files should already be available.
If this script is usable I will need to modify the server code so it puts those JSON files in a jsonld directory rather than dist/jsonld as it does now.


prakhyatox:
The script supports the following build modes:

* When the environment variable is set to false, it skips the full build and generates index.html files only for static URLs.
* When the environment variable is set to true, it performs the full build, including generating index.html files for approximately 8,000 records and 5,000 organisations.
* If the API server is unavailable, the build can still complete by using API data saved from the previous successful build.

The main purpose of the script is to generate static index.html files for records, organisations, and other pages. These files are important for SEO because search engine crawlers generally rely on the HTML generated during the build.

The SEO metadata is generated from the page data at build time. Therefore, changes to the source code, environment variables, or record data will not appear on the live site until a new build is completed and deployed.

The simplified script does not currently perform this full generation process. Even if the record data is read from existing JSON-LD files instead of being fetched from the API, a new build is still required to update the generated HTML used by search engines.

Using existing JSON files for production builds should be possible. However, we will need to agree on their location. If the files will be stored in a jsonld directory instead of dist/jsonld, the server code and build script will need to be updated accordingly.

knirirr:
> The simplified script does not currently perform this full generation process. Even if the record data is read from existing JSON-LD files instead of being fetched from the API, a new build is still required to update the generated HTML used by search engines.

Please feel free to modify it as you feel necessary. The current script must be changed anyway as it cannot reflect changes in source code or the .env file. 

> Using existing JSON files for production builds should be possible. However, we will need to agree on their location. If the files will be stored in a jsonld directory instead of dist/jsonld, the server code and build script will need to be updated accordingly.

I would suggest simply a top level fairsharing.github.io/jsonld directory. If you are happy with this I'll make the change.

prakhyatox:
1. I have updated the script and now it is reflecting the latest changes.
2. I have made changes to jsonld path also reading it from top level directory

Updated script can be found in https://github.com/FAIRsharing/fairsharing.github.io/pull/2830/changes

knirirr:
Won't be needed.

Issue #2755, opened 2026-04-28
Add documentation chatbot to preview.fairsharing.org


allysonlister:
Please can we add the IntelliDoc chatbot created by the AICC and Alok to our preview site for wider testing?

IntelliDoc Webpage URL:
https://aicc.uksouth.cloudapp.azure.com/login 

if you need the login details please contact me on zulip.

Ask Alok if you are unsure how to proceed; I thought he mentioned that there was some code that could be copied from the IntelliDoc project page to embed within our site.

fyi @ramGranell

allysonlister:
fyi @knirirr @prakhyatox this has gained a high priority.

Primarily have a think about the size - what happens 'by default' using IntelliDoc? We don't want the chat window to take over the whole page, but rather be about the size of what other websites have.

PR #2834, opened 2026-07-15
Fiona chatbot

prakhyatox:
TIcket: https://github.com/FAIRsharing/fairsharing.github.io/issues/2755

Fiona chatbot is implemented


knirirr (approved):

allysonlister (approved):
My preference is to do a soft launch (no socials) and make it public on prod - if you agree, please do so.

Issue #2835, opened 2026-07-16
Issue with retrieving zenodo DOI in record edit interface


allysonlister:
I am attempting to retrieve publication details for DOI 10.5281/zenodo.20703769 from zenodo via any record edit interface and it's just spinning/hanging - @knirirr has noted that this is a client side issue:

It's a client-side bug. The server is responding correctly.

TypeError: can't access property 0, t.created["date-parts"] is undefined
    getDOI EditPublications.vue:593
    2 EditPublications.vue:153

PR #2836, opened 2026-07-16
Dealt with bug in #2835, fixed test timeouts.

knirirr:
Deals with 2835 and fixes a test timeout.


allysonlister:
Thanks - this looks like it's working just fine for the DOI from zenodo now.

allysonlister (approved):

Issue #2849, opened 2026-07-28
New Educational tile


allysonlister:
Please add this tile to the /educational page.

<img width="1024" height="768" alt="Image" src="https://github.com/user-attachments/assets/60f9bd5c-1c81-4591-b800-94d38f207c89" />

associated DOI is https://doi.org/10.5281/zenodo.21628854 (for the "copy" icon) and the direct link for download is https://zenodo.org/records/21628855/files/v1.0%20-%20FAIRsharing%20behind%20the%20scenes_The%20FAIRness%20of%20FAIRsharing.pdf?download=1

allysonlister:
Forgot to say where. Please add to the bottom row, at the end.

thanks!

Issue #2843, opened 2026-07-23
Remove hyperlink to stakeholder


allysonlister:
<img width="1368" height="147" alt="Image" src="https://github.com/user-attachments/assets/de868676-da94-4954-b15a-b2c4f787da62" />
Please remove the url of stakeholder from https://fairsharing.org/communities

PR #2850, opened 2026-07-28
Stakeholder link 2843

knirirr:
This deletes an extraneous link in #2843 and also adds a new image/tile as per #2849 
Once this has been reviewed I will deploy all pending changes recently merged to dev.


allysonlister (approved):
working as expected, thanks

